Issue #3, opened 2020-09-18
Review multibase vs base58 key representation


OR13:
@dlongley I recall you may have objected to the idea of `publicKeyMultibase` vs `publicKeyBase58`.

Would love to hear your thoughts.

My initial thoughts are that `publicKeyMultibase` might be easier to align with IPLD in the future.

peacekeeper:
Is [Example 4](https://w3c-ccg.github.io/lds-ed25519-2020/#example-4) correct?

If you have

```
"publicKeyBase58": "dbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1",
```

then shouldn't that be equivalent to

```
"publicKeyMultibase": "zdbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1",
```

?

But the example currently seems to show two completely different key pairs.

dlongley:
I don't mind `publicKeyMultibase` ... I'm just worried about people needing to support all your base. If we do `publicKeyMultibase`, someone is going to start using things [other than base58](https://github.com/multiformats/multibase#multibase-table) and this will lead to annoying interop problems. We need some kind of brakes on what gets supported.

For example -- we could say that the linked data key type can say which bases are supported (ideally just one).

OR13:
Agree, I think throwing an error in the library when an unsupported base is detected is the best solution. 

And let the signature suite itself define the supported bases, so that support for a suite automatically causes libraries to support popular bases.

peacekeeper:
But again, [Example 4](https://w3c-ccg.github.io/lds-ed25519-2020/#example-4) shows two different keys, even though they both have the id `https://example.com/issuer/123#key-0`, right?

If someone confirms that, I could try to fix the example(s).

OR13:
IIRC they are supposed to be the same key encoded 2 different ways, or that was my intention.

the `z` tells you its `base58btc` multicodec. My intention was to provide examples that showed how to upgrade an `Ed25519VerificationKey2018`.

It looks like I did the math wrong.

OR13:
its also possible that the prefixing shifted the base58 encoding of the binary, and thats why the character strings look different.

OR13:
the desired behavior is to show both encodings for the same binary key material.

clehner:
Hi @OR13, @peacekeeper,
I notice this encoding issue also.

[...]

PR #10, opened 2021-06-04
Update examples

clehner:
As discussed in #3, the multibase-encoded values in examples were not right but were encoded with multibase twice. Fortunately, I was able to verify the example verifiable credential `vc_0` after decoding the `proofValue` and the issuer `publicKeyMultibase` values. The verifiable presentation `vp_0` I was not able to verify: it may be stuck with the incorrectly-encoded VC proof value; also, it has an `id` property which is not a URI, and I don't know the base IRI to resolve it as a relative URI reference. So I change the VP id to a UUID, and regenerate the VP proof. I also change the issuer document to use `verificationMethod`, since `publicKey` was deprecated and removed from the context file: https://github.com/w3c/did-spec-registries/pull/277/files#r602970448. And I change the contexts to use the live context file for this repo, similarly to how it is done in <https://w3c-ccg.github.io/lds-jws2020/>.

To verify the example VC and generate the new example VP proof, I use the following code: https://github.com/spruceid/ssi/commit/1c90a9b7d5501aa67c0a6d2d7b37981eaaf1d78d

### DIFF
--- a/index.html
+++ b/index.html
@@ -383,7 +383,7 @@
               "id": "https://example.com/issuer/123#key-0",
               "type": "Ed25519VerificationKey2020",
               "controller": "https://example.com/issuer/123",
-              "publicKeyMultibase": "zgo4sNiXwJTbeJDWZLXVn9uTnRwgFHFxcgDePvEC9TiTYgRpG7q1p5s7yRAic"
+              "publicKeyMultibase": "zdbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1"
             }
           </pre>
 
@@ -401,7 +401,7 @@
                   "id": "#key-0",
                   "type": "Ed25519VerificationKey2020",
                   "controller": "did:example:123",
-                  "publicKeyMultibase": "zgo4sNiXwJTbeJDWZLXVn9uTnRwgFHFxcgDePvEC9TiTYgRpG7q1p5s7yRAic"
+                  "publicKeyMultibase": "zdbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1"
                 }
               ],
               "authentication": [
@@ -488,8 +488,8 @@
             "id": "https://example.com/issuer/123#key-0",
             "type": "Ed25519KeyPair2020",
             "controller": "https://example.com/issuer/123",
-            "publicKeyMultibase": "zgo4sNiXwJTbeJDWZLXVn9uTnRwgFHFxcgDePvEC9TiTYgRpG7q1p5s7yRAic",
-            "privateKeyMultibase": "z5LgFTdoZ3hjXbijBYC8qBEEZ1oibq1AKVeSeuw84zgYzUNY2rXKT8xLU41XpMN124wVr26axAPBPiNnR9dvyY7KA9QCmTrxT6yCCJxfS6U5iEBhzfbWqPSazGJ"
+            "publicKeyMultibase": "zdbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1",
+            "privateKeyMultibase": "z47QbyJEDqmHTzsdg8xzqXD8gqKuLufYRrKWTmB7eAaWHG2EAsQ2GUyqRqWWYT15dGuag52Sf3j4hs2mu7w52mgps"
           }
         }              
       </pre>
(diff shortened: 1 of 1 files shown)


OR13 (approved):
@dmitrizagidulin @msporny can you review this? We are not actively using 2020 yet.

index.html
```diff
@@ -383,7 +383,7 @@ <h3>Ed25519VerificationKey2020</h3>
               "id": "https://example.com/issuer/123#key-0",
               "type": "Ed25519VerificationKey2020",
               "controller": "https://example.com/issuer/123",
-              "publicKeyMultibase": "zgo4sNiXwJTbeJDWZLXVn9uTnRwgFHFxcgDePvEC9TiTYgRpG7q1p5s7yRAic"
+              "publicKeyMultibase": "zdbDmZLTWuEYYZNHFLKLoRkEX4sZykkSLNQLXvMUyMB1"
```
msporny, line 386: This looks like it's off by 1-2 bytes (should be 45-46 bytes, I'm seeing 44 bytes)... possibly due to varint encoding for the `0xed` value? A `publicKeyMultibase` value for `Ed25519VerificationKey2020` should start with a 'z' and then a `0xed01` byte header (which is the `0xed` multicodec value encoded as a varint). I'm doing this rushed and from memory, so I might be wrong here.

@dmitrizagidulin it looks like this bug exists in our code as well: https://github.com/digitalbazaar/ed25519-signature-2020/blob/main/test/mock-data.js#L11

... or was there some decision made somewhere that we're not doing multicodec encoding of multibase public keys anymore? /cc @dlongley

msporny: Yeah, this looks totally wrong: https://github.com/digitalbazaar/ed25519-verification-key-2020/blob/main/lib/Ed25519VerificationKey2020.js#L104

dlongley: It needs to be fixed to include the multikey prefix.

clehner: Addressing in #11

OR13: shouldn't `publicKeyMultibase` match a did:key, for example:

`did:key:z6Mkn9p3p3MEKekdhKQ7SYZ91LjeSUgiYkzin4VPt6BeHNj8`

`z6Mkn9p3p3MEKekdhKQ7SYZ91LjeSUgiYkzin4VPt6BeHNj8` is a base58 encoded ed25519 public key.

for existing key types, I would expect them to be resolvable when prefixed with did:key

OR13: highly possible I have just not looked at this in a while and am confused, but I just spent all weekend parsing public  key bytes from did key fingerprints... and it would be a shame for that to be somehow not the same as this.

msporny: Yes `publicKeyMultibase` should match a `did:key` (what was one of the goals of using `publicKeyMultibase` over base58btc) -- I know DB's implementation is broken in this regard and we're fixing it for the 2020 suite.

OR13: This means that there are a lists of mostly usable examples here: https://github.com/w3c-ccg/did-method-key/tree/master/test-vectors

hidden inside of the didDocument.id property.

but beware that bls12381 as a format that has 2 public keys in it, because its a pairing friendly curve.

OR13, line 386:
```suggestion
              "publicKeyMultibase": "z6MkiTBz1ymuepAQ4HEHYSF1H8quG5GLVVQR3djdX3mDooWp"
```

https://github.com/w3c-ccg/did-method-key/blob/master/test-vectors/ed25519-x25519.json#L2

clehner: The `z6Mk` prefix for Ed25519 public key (like in [did:key](https://w3c-ccg.github.io/did-method-key/#ed25519-x25519)) can be seen in #11

OR13 (changes requested):
publicKeyMultibase examples are not correct, see #11.

index.html
clehner: I got `z6Mkf5rGMoatrSj1f4CyvuHBeXJELe9RPdzo2PKGNCKVtZxP` for this keypair (`#key-0`) in multibase/multicodec format; in #11: https://github.com/w3c-ccg/lds-ed25519-2020/pull/11/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R520

OR13: yeah, the examples should probably be updated consistently, I am just pointing out the obvious issue with it needing to look like: `z6Mk`

clehner:
#11 is superseding this PR.

PR #12, opened 2021-07-22
Ed25519 uses SHA-512, not SHA-256

peacekeeper:


index.html
```diff
@@ -314,8 +314,8 @@ <h2>Suite Definition</h2>
           </tr>
           <tr>
             <td><a>message digest algorithm</a></td>
-            <td>SHA-256</td>
-            <td>[[RFC4634]]</td>
+            <td>SHA-512</td>
+            <td>[[RFC6234]]</td>
```
msporny, line 318: While it is true that the general version of ed25519 (not prehash or ctx) uses sha-512 internally, the value here is not describing that value.

The value here is describing the hashing algorithm that is used to hash the output of the canonicalization algorithm, which is SHA2-256, before the hash output is sent to the signing algorithm:

https://github.com/digitalbazaar/jsonld-signatures/blob/main/lib/suites/LinkedDataSignature.js#L241

This is one of those badly explained things in the current Linked Data Proofs/Signatures specs and will hopefully be clarified in the upcoming LDS work that will be taken up in the VCWG.

msporny (changes requested):
The spec is correct as is... explanation above.

index.html
peacekeeper: Right.. There are also crypto suites such as [JcsEd25519Signature2020](https://github.com/decentralized-identity/JcsEd25519Signature2020) which use different canonicalization, but then hashing as part of the signing algorithm is still done in the same way.

peacekeeper:
Closing per explanation above.

PR #20, opened 2022-12-01
Add Dmitri Zagidulin as editor.

dmitrizagidulin:
I would like to volunteer to serve as editor of this spec.


tplooker (approved):
Thanks!

PR #23, opened 2023-01-31
Added test vectors for the Ed25519Signature2020 case.

Wind4Greg:
Added a set of test vectors for the `"type": "Ed25519Signature2020"` case to Appendix A. Let me know if this is the approach we'd like to use.

Cheers
Greg B.

### DIFF
--- a/index.html
+++ b/index.html
@@ -814,150 +814,161 @@ unlinkability. If signatures are re-used, they can be used as correlatable data.
 
     <section class="appendix">
       <h2>Test Vectors</h2>
+      <section>
+        <h3>Representation: Ed25519Signature2020</h3>
+        <p>
+The signer needs to generate a private/public key pair with the private key used 
+for signing and the public key made available for verification. Below the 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -814,150 +814,161 @@ <h2>Privacy Considerations</h2>
...
+    "proofPurpose": "assertionMethod",
+    "proofValue": "z2RczMj342tVhAjgjEPV4TeHbi2ggnTRKTc5BFQCgaWJ3nhcg5HgCeC2eV4Lc1fYdhfoLyPjxoq4BtqrsyNvxZ8nE"
   }
 }
       </pre>
+
+    </section>
+      <section>
+        <h3>Representation: DataIntegrityProof</h3>
+        <p>To be written...</p>
```
OR13, line 970: maybe add a link to an issue.

msporny: Agree with @OR13, this feels out of place. Either link to an issue, or remove it from the PR (since, presumably, this section will appear in a future PR).

Wind4Greg: Agree. Will update and reissue.

PR #26, opened 2023-02-25
Add json-eddsa-2022 cryptosuite.

msporny:
Per [the discussion during the VCWG in-person meeting in Miami](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-02-15-vcwg#section1-2), the `jws-2020` cryptosuite was going to support JCS, but since EdDSA and ECDSA cryptosuites were going to also concretely support the feature, a decision was made to abandon the jws-2020 cryptosuite and merge the JCS featureset into the Data Integrity cryptosuites. 

This PR concretely adds the JCS functionality into the EdDSA Cryptosuite.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/pull/26.html" title="Last updated on Apr 4, 2023, 2:29 PM UTC (af98f82)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/26/366354c...af98f82.html" title="Last updated on Apr 4, 2023, 2:29 PM UTC (af98f82)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -813,10 +813,24 @@ Return <var>verificationResult</var> as the <em>verification result</em>.
       </section>
 
       <section>
-        <h3>json-eddsa-2022</h3>
+        <h3>jcs-eddsa-2022</h3>
+
+          <p class="issue" title="Cryptosuite naming convention is disputed">
+The naming convention utilized by this cryptosuite is disputed. An alternative
+of `json-eddsa-2022` was originally suggested for this cryptography suite to
+convey that it is a cryptography suite for securing JSON data utilizing the
+Twisted Edwards Curve Digital Signature Algorithm. The counter-argument to
+the original proposal was that expressing the canonicalization mechanism in
+the cryptosuite string clearly conveys to a developer that the thing that
+differentiates this cryptosuite from the `eddsa-2022` one is the use of
+JSON Canonicalization Scheme [[RFC8785]]. Other options include
+`"cryptosuite": "json-sign-2022"`, and `"cryptosuite": "json-2022"`. This
+topic is <a href="https://github.com/w3c/vc-data-integrity/issues/38">
+currently being debated in the Data Integrity work item.</a>.
+          </p>
 
           <p>
-The `json-eddsa-2022` cryptographic suite takes an input document, canonicalizes
+The `jcs-eddsa-2022` cryptographic suite takes an input document, canonicalizes
 the document using the JSON Canonicalization Scheme [[RFC8785]], and then
 cryptographically hashes and signs the output resulting in the production of a
 data integrity proof. The algorithms for this cryptographic suite are the
@@ -833,7 +847,7 @@ and step <strong>2)</strong> are replaced by the following text:
             <li>
 If <var>options</var>.<var>type</var> is not set to the string
 `DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
-set to the string `json-eddsa-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
+set to the string `jcs-eddsa-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
 raised.
             </li>
             <li>
--- a/index.html
+++ b/index.html
@@ -595,7 +595,7 @@ into a transformed document that is ready to be provided as input to the
 hashing algorithm in Section <a href="#hashing-eddsa-2022"></a>.
           </p>
 
-
+          <p>
 Required inputs to this algorithm are an
 <a href="https://w3c.github.io/vc-data-integrity/#dfn-unsecured-data-document">
 unsecured data document</a> (<var>unsecuredDocument</var>) and
@@ -810,6 +810,48 @@ Return <var>verificationResult</var> as the <em>verification result</em>.
           </ol>
 
         </section>
+      </section>
+
+      <section>
+        <h3>json-eddsa-2022</h3>
+
+          <p>
+The `json-eddsa-2022` cryptographic suite takes an input document, canonicalizes
+the document using the JSON Canonicalization Scheme [[RFC8785]], and then
+cryptographically hashes and signs the output resulting in the production of a
+data integrity proof. The algorithms for this cryptographic suite are the
+same as the ones in Section <a href="#eddsa-2022"></a> except for the following
+modifications:
+          </p>
+
+          <p>
+In Section <a href="#transformation-eddsa-2022"></a>, step <strong>1)</strong>
+and step <strong>2)</strong> are replaced by the following text:
+          </p>
+
+          <ol class="algorithm">
+            <li>
+If <var>options</var>.<var>type</var> is not set to the string
+`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
+set to the string `json-eddsa-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
+raised.
+            </li>
+            <li>
+Let <var>canonicalDocument</var> be the result of applying the
+JSON Canonicalization Scheme [[RFC8785]] to the <var>unsecuredDocument</var>.
+            </li>
+          </ol>
+
+          <p>
+In Section <a href="#proof-configuration-eddsa-2022"></a>, step
+<strong>4)</strong> is replaced by the following text:
+          </p>
+
+          <p style="padding-left: 2em;">
+<strong>4)</strong> If <var>options</var>.<var>type</var> is not set to
+`DataIntegrityProof` and <var>proofConfig</var>.<var>cryptosuite</var> is not
+set to `json-eddsa-2020`, an `INVALID_PROOF_CONFIGURATION` error MUST be raised.
+          </p>
       </section>
 
       <section>
--- a/index.html
+++ b/index.html
@@ -53,6 +53,14 @@
           companyURL: "https://digitalcredentials.mit.edu/"
         }],
 
+        authors: [{
+          name: "Dave Longley", url: "http://digitalbazaar.com/",
+          company: "Digital Bazaar", companyURL: "http://digitalbazaar.com/"
+        }, {
+          name: "Manu Sporny", url: "https://www.linkedin.com/in/manusporny/",
+          company: "Digital Bazaar", companyURL: "https://digitalbazaar.com/"
+        }],
+
         github: "https://github.com/w3c/vc-di-eddsa/",
 
         // URI of the patent status for this WG, for Rec-track documents


index.html
```diff
@@ -807,6 +807,48 @@ <h3>Proof Verification (eddsa-2022)</h3>
         </section>
       </section>
 
+      <section>
+        <h3>json-eddsa-2022</h3>
```
dlongley: I wonder if this should be `jcs-eddsa-2022` instead. Otherwise, the rest of this is a good start, IMO.

dlongley (approved):

index.html
msporny: Yes, I had started w/ that and then became concerned that most developers won't know or care about what "JCS" is... even "eddsa" is a stretch -- most developers don't know anything about the details of how "EdDSA" works, or what parameters they should be using for their application... For this particular cryptosuite, they're probably just looking for a way to sign their JSON and perhaps the name should reflect that. I was wondering if we should go further and just have aliases like `"cryptosuite": "json-2023"` instead... 'cause that might be more meaningful to a developer than `eddsa`.

OR13: I think it should say `jcs` if JCS is required.

OR13:
```suggestion
        <h3>jcs-eddsa-2022</h3>
```

OR13 (changes requested):
JCS should be in the short name

index.html
msporny: -1 to JCS being in the name for at least the following two reasons:

* We don't state the canonicalization algorithm name, on purpose, in any of the other cryptosuites. That is... it's not `urdna-eddsa-2022` nor is it `UrndaEd25519Signature2020`.
* Even if we state it, developers are unlikely to know or care about `jcs` being used under the hood.

@dlongley @OR13, is there an alternative name, that will help developers understand that this signature works on "JSON documents", that either of you would be willing to use other than `jcs`?

OR13: `UrndaEd25519Signature2020.` <- this would be a much more "honest" name...

Callings JCS "JSON" is very misleading.

iherman:
The issue was discussed in a [meeting](https://www.w3.org/2017/vc/WG/Meetings/Minutes/2023-03-08-vcwg#section2-5) on 2023-03-08

- no resolutions were taken

<details><summary><i>View the transcript</i></summary>

#### 2.5. Add json-eddsa-2022 cryptosuite. (pr vc-di-eddsa#26)

_See github pull request [vc-di-eddsa#26](https://github.com/w3c/vc-di-eddsa/pull/26)._




**Manu Sporny:** for EDDSA, there are 2 PRs, one just updates test vectors, the other one is naming things.  
… so if you want to get involved in bikeshedding, on what to call the crypto suite that uses JCS for canon + signs with Eddsa, comment on issue 26.  

</details>

TallTed (commented):
LGTM...

I do wonder whether developers working with these specs will find it sufficiently easy to blend the algorithm steps across documents. I sympathize with the desire to minimize maintenance of the same text in multiple docs, but I wonder whether it's possible to refactor the algorithms and docs sufficiently to use something like an `include` to reconstruct the entire algorithm such that it can be read on one page?

msporny:
> I do wonder whether developers working with these specs will find it sufficiently easy to blend the algorithm steps across documents.

Yes, I agree, that's a problem. I've not reworked the specifications multiple times to try and make it easier. I don't think we get to the next stage without doing what you say below...

> I sympathize with the desire to minimize maintenance of the same text in multiple docs, but I wonder whether it's possible to refactor the algorithms and docs sufficiently to use something like an `include` to reconstruct the entire algorithm such that it can be read on one page?

Yep, I agree, that's probably the right end state. It requires some ReSpec code to be written such that you can pass local document sections in the `include` construct. If we do that, then you just have to look at the cryptosuite of interest instead of trying to piece together the algorithm across multiple specs. "A templating language to provide English as a programming language so that developers can write programs in real programming languages." -- hard to get much more meta than that. :P

index.html
msporny: I don't think we're going for "honesty" here... or technical accuracy. 

"json" does describe what a non-expert developer might want... "I want to sign my JSON data... oh 'json-SOMETHING', that looks good, I'll use that". That's the sort of thought process that I'd hope we're going for. Not a string of acronyms where they have to go and look up what each acronym means (and even then, they don't have the expertise to choose between something like JCS vs. URDNA2015).

We're going for something that a non-expert developer can easily understand (because they're almost surely not going to be able to understand all the parameters that are in play).

Like: "selective-disclosure-2023" is probably better than "sdjwt-2023" or "bbsdsa-2023".

I'm cringing at the fact that we have "eddsa" vs. "ecdsa" at present... one letter difference (granted, they're both elliptic curves, where one of them is believed to possibly have been tampered with by the NSA and the other one is provably untampered with relation to the curve parameters).

Perhaps "nist-2019" vs. "not-pwnd-by-the-nsa-2019" would be better?

I'll note that even the HPKE folks picked something like `HPKE-V1-BASE` or something, somewhere in their specs.

We need to get away from all the "cryptobabble" for non-expert developers. So, while I'm not wed to using the term 'json'... I don't want us to fall into the cryptobabble trap w/ our cryptosuite names.

OR13: > I'm cringing at the fact that we have "eddsa" vs. "ecdsa" at present.

So am I... what happens if you need to add support for new crypto?

All this can be hidden in the "suite name" or in a "parameter registry"... or a combo.

Why not use `DigitalSignature2023`, and then in the registry entry for the suite, explain that is EdDSA with URDNA2015 and sha256?

Or name it `DigitalSignature-EdDSA-URNDA2015-SHA-256-2023`

The name doesn't need to expose all the parameters, if you learn all of them from reading its definition in a registry...

The current approach, uses a mixture of the suite name and parameters... If you want to be like HPKE, you would create a register for each parameter class, and then call the signature class something like:

DataIntegrityProof which would define support for all the predicates necessary to represent the registry entries that exist, for example:

```json
"type": "DataIntegrityProof",
"canonicalization": "JCS",
"digest": "SHA-256",
"cipherSuite": "EdDSA",
"proofValueMultibase": "base64 \ base58 \ base32 \ baseEmoji ..."
```

Or you could do this:

```json
"type": "DataIntegrityProof2023",
"proofValueMultibase": "base64 \ base58 \ base32 \ baseEmoji ..."
```

And in the registry say.... "DataIntegrityProof2023" is "JCS" + "SHA-256" + "EdDSA"

silverpill:
@msporny Hi, I'm the author of [FEP-c390](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-c390.md) document. I received a couple of emails from you but I'm not sure if my reply was properly delivered because my address belongs to a throwaway email provider...

Thanks for making this PR! Indeed, FEP-c390 and [FEP-8b32](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-8b32.md) (another proposal based on Data Integrity spec) currently recommend using JCS, so it's good to have support for this algorithm. However, algorithm recommendations may change as we gather more feedback from implementers. It is quite possible that we'll end up using URDNA2015 because it has more libraries for various programming languages (in particular, Mastodon is written in Ruby; I haven't found any JCS library for Ruby, but there is an RDF library and Mastodon already uses it for Linked Data signatures).

Do similar specifications exist for other signature algorithms, like RSA? I had some difficulty finding information about cryptosuites, see this issue: https://github.com/w3c/vc-data-integrity/issues/84

>jcs-eddsa-2022 vs json-eddsa-2022

As a developer, I would prefer a more descriptive name: `jcs-eddsa-2022`. Descriptive names improve readability and reading JSON documents is something that we do a lot, especially when trying to resolve interoperability issues.

OR13:
@silverpill thanks for commenting on the PR.

> As a developer, I would prefer a more descriptive name: `jcs-eddsa-2022`

+1 to this.

msporny:
@silverpill wrote:
> @msporny Hi, I'm the author of [FEP-c390](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-c390.md) document. I received a couple of emails from you 

Hooray, you're here! :) Hi -- been trying to get a hold of you for a while.

> but I'm not sure if my reply was properly delivered because my address belongs to a throwaway email provider...

Uh oh, yeah, I didn't receive a single email from you. Our company recently had to abandon running our own email servers because Google was blocking delivery to Gmail inboxes (even though we ticked all the DMARC, IP reputation, DNS SEC checkboxes). We eventually gave up after important emails were getting sporadically blocked. 

In any case, we have your attention on the PR that I was hoping to direct you towards given your work on FEP-c390.

> Thanks for making this PR! Indeed, FEP-c390 and [FEP-8b32](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-8b32.md) (another proposal based on Data Integrity spec) currently recommend using JCS, so it's good to have support for this algorithm.

Good, FEP-c390 was one of the primary drivers for this PR.

> However, algorithm recommendations may change as we gather more feedback from implementers. It is quite possible that we'll end up using URDNA2015 because it has more libraries for various programming languages (in particular, Mastodon is written in Ruby; I haven't found any JCS library for Ruby, but there is an RDF library and Mastodon already uses it for Linked Data signatures).

Yeah, understood. I guess the good news is that the ActivityPub community will have both options open to you.

What we don't know yet, and which you might be able to provide some input on, is how much implementations conform to these statements today:

https://www.w3.org/TR/activitystreams-core/#jsonld

and here:

https://www.w3.org/TR/activitypub/#obj

That is, given that we have five years of implementation experience today, would implementers have a preference using URDNA2015 vs. JCS? or, do they just want their data signed and they don't really care about the canonicalization algorithm? (I'm specifically NOT trying to bias you one way or the other by giving you the benefits and drawbacks of each).

> Do similar specifications exist for other signature algorithms, like RSA? I had some difficulty finding information about cryptosuites, see this issue: [w3c/vc-data-integrity#84](https://github.com/w3c/vc-data-integrity/issues/84)

I'll answer you in the issue in a bit, but yes, there are other specs for other signature algorithms:

https://w3c.github.io/vc-di-eddsa/
https://w3c-ccg.github.io/vc-di-ecdsa/

and some older ones that are going to get big changes applied to them (so, don't copy anything in them, or presume they're right, please):

https://w3c-ccg.github.io/ld-cryptosuite-registry/#signature-suites

We're going to merge the above registry over to here shortly:

https://w3c.github.io/vc-specs-dir/#proof

> > jcs-eddsa-2022 vs json-eddsa-2022
> 
> As a developer, I would prefer a more descriptive name: `jcs-eddsa-2022`. Descriptive names improve readability and reading JSON documents is something that we do a lot, especially when trying to resolve interoperability issues.

A couple of questions for you:

Do you think most developers (that is, not you, but people with less skill than you have that are fediverse implementers) will understand the difference between `jcs` and `urdna2015`? Do you think they are capable of understanding how using SHA-2-256, SHA-3-256, and SHAKE256 will affect their application?

index.html
msporny: @OR13 wrote:
> And in the registry say.... "DataIntegrityProof2023" is "JCS" + "SHA-256" + "EdDSA"

Yes, exactly -- that's more or less where I hope we'll end up (as a WG).

We use a cryptosuite name that's easy for devs, without any/much security training, to pick and use as they're copy-pasting their way through an implementation (which is the default mode of operation for most developers). We can elaborate what that term means in a directory or the spec.

> > I'm cringing at the fact that we have "eddsa" vs. "ecdsa" at present.
> 
> So am I... what happens if you need to add support for new crypto?

You use a new base name and/or you iterate the version/date?

> All this can be hidden in the "suite name" or in a "parameter registry"... or a combo.

Yep, exactly.

> Why not use `DigitalSignature2023`, and then in the registry entry for the suite, explain that is EdDSA with URDNA2015 and sha256?

Yes, that's exactly what I'm trying to say.

> Or name it `DigitalSignature-EdDSA-URNDA2015-SHA-256-2023`

No, thanks... that's cryptobabble to most developers. 

That and `DigitalSignature-EdDSA-URNDA2015-MD5` would probably look so similar to most devs that mistakes would be made. In fact, that is exactly how mistakes are made today during algorithm selection with TLS <1.3 and `alg=none` and using `HS256` instead of `ES256`. 

We've gotta stop enabling devs to pick dangerous parameters.

> The name doesn't need to expose all the parameters, if you learn all of them from reading its definition in a registry...

Yes, exactly.

> The current approach, uses a mixture of the suite name and parameters... 

How so? The current (primary) trajectory we're on is: `ecdsa-2019`, `eddsa-2022`, and `bbs-2023` -- none of those let devs pick parameters. `jcs` would be the first time where we introduce this... and, arguably `json` isn't a whole lot better. A better name could be something like `"cryptosuite": "simple-json-2023"` (which is still clearly not great)... maybe  `"cryptosuite": "json-2023"`... or  `"cryptosuite": "default-2023"`?

> DataIntegrityProof which would define support for all the predicates necessary to represent the registry entries that exist, for example:
> 
> ```json
> "type": "DataIntegrityProof",
> "canonicalization": "JCS",
> "digest": "SHA-256",
> "cipherSuite": "EdDSA",
> "proofValueMultibase": "base64 \ base58 \ base32 \ baseEmoji ..."
> ```

The above suggests that developers can pick and chose the parameters, which we don't want them to be able to do.

> Or you could do this:
> 
> ```json
> "type": "DataIntegrityProof2023",
> "proofValueMultibase": "base64 \ base58 \ base32 \ baseEmoji ..."
> ```

This would put us back into the "cryptosuite type proliferation" that you argued strongly against, and is why we went to `DataIntegrityProof` with a `cryptosuite` parameter in the first place.

That said, feels like we're converging on an approach. Given that we want to support EdDSA, ECDSA, some kind of selective disclosure (in time), and some kind of unlinkable signature (in time)... and given that the parameters for those algorithms are going to change over time... whats a unified naming scheme we can apply to these cryptosuite names?

silverpill:
@msporny 

>What we don't know yet, and which you might be able to provide some input on, is how much implementations conform to these statements today:

Most implementations treat objects and activities as plain JSON. Everyone includes `https://www.w3.org/ns/activitystreams` context, but other contexts could be missing or incorrect.

>That is, given that we have five years of implementation experience today, would implementers have a preference using URDNA2015 vs. JCS? or, do they just want their data signed and they don't really care about the canonicalization algorithm? (I'm specifically NOT trying to bias you one way or the other by giving you the benefits and drawbacks of each).

I think they just want to sign their data. Or, more precisely, they *may* want to sign data, but if the process is too complicated, they will rather do something else. This is the reason why I initially picked JCS as a recommended canonicalization algorithm for my proposal: it looks less daunting.

>Do you think most developers (that is, not you, but people with less skill than you have that are fediverse implementers) will understand the difference between jcs and urdna2015?

I think most people will follow "best practices" and won't spend much time trying to understand the nature of these algorithms. What constitutes best practices is a big question, of course, and right now this is mostly "what Mastodon is doing", but the purpose of Fediverse Enhancement Proposals is to provide guidance. If FEP recommends using JCS, EdDSA and SHA-256, I think developers will grab necessary libraries, implement integrity proofs, test if interoperability is not broken and forget about it.

>Do you think they are capable of understanding how using SHA-2-256, SHA-3-256, and SHAKE256 will affect their application?

Probably not. To be honest, I don't understand that either :)
I would pick SHA-2-256 unless there's a good reason to do otherwise.

msporny:
> > What we don't know yet, and which you might be able to provide some input on, is how much implementations conform to these statements today:
> 
> Most implementations treat objects and activities as plain JSON. Everyone includes `https://www.w3.org/ns/activitystreams` context, but other contexts could be missing or incorrect.

Yes, and that's what we predicted might happen while those specifications were being put together, and is why the activitystreams context has a default vocabulary set to `_:`. That makes any undefined term a document-scoped blank node property in JSON-LD, but drops it when you go to RDF / N-Quads (which you would if using URDNA2015). So, that's one signal that the fediverse, for legacy content, might be better off using JCS.

> I think they just want to sign their data. Or, more precisely, they _may_ want to sign data, but if the process is too complicated, they will rather do something else. This is the reason why I initially picked JCS as a recommended canonicalization algorithm for my proposal: it looks less daunting.

Yep, JCS is less daunting. Granted, you lose some functionality if you use it vs. using URDNA2015 (like the ability to address information in a way that enables more advanced cryptographic mechanisms to be used like selective disclosure and unlinkable digital signatures) , but that loss is probably not important for the Fediverse use cases. You're not dealing w/ financial transactions, or extensions that have grave financial or political outcomes if some subset of the message were not to be signed.

So, I agree that JCS is probably the right call for the Fediverse *if* most devs aren't including their JSON-LD extension Contexts (are there any? If so, how many?) OR if it's not easily detectable which JSON-LD extensions are being used (is it possible to detect extensions and map them to JSON-LD Contexts?). Is there a registry of some kind for Fediverse extensions outside of FEPs?

> > Do you think most developers (that is, not you, but people with less skill than you have that are fediverse implementers) will understand the difference between jcs and urdna2015?
> 
> I think most people will follow "best practices" and won't spend much time trying to understand the nature of these algorithms. 

Yes, and that's exactly why giving fediverse devs the ability to do parameter selection for cryptographic algorithms is almost certainly a bad idea.

> What constitutes best practices is a big question, of course, and right now this is mostly "what Mastodon is doing", but the purpose of Fediverse Enhancement Proposals is to provide guidance. If FEP recommends using JCS, EdDSA and SHA-256, I think developers will grab necessary libraries, implement integrity proofs, test if interoperability is not broken and forget about it.

Yep, that's aligned w/ the mental model we believe fediverse developers will have, based on how we've seen them use this stuff in the past.

> > Do you think they are capable of understanding how using SHA-2-256, SHA-3-256, and SHAKE256 will affect their application?
> 
> Probably not. To be honest, I don't understand that either :) I would pick SHA-2-256 unless there's a good reason to do otherwise.

You'd have made the right choice for this year, but not understanding why you've made the right choice is the concern here.

So, going back to "How do we name the data integrity cryptosuites?", we have two general directions:

1. Use a cryptosuite name like `default-2023` or `fediverse-2023` ... which will be explained to use "JCS + EdDSA w/ Curve25519" in some registry/spec somewhere, or
2. Give the developers parameter options in the cryptosuite string that allow them to pick parameters: "[jcs|urdna2015]-[eddsa-ecdsa]-[sha2-256|sha3-256|shake256]-[p256|p384|p521|curve25519|golidlocks448]"

Of those two options, I expect developers to footgun themselves far more often with option 2 than option 1... adding `jcs` to the name of the cryptosuite feels like an argument for option 2.

silverpill:
>So, I agree that JCS is probably the right call for the Fediverse if most devs aren't including their JSON-LD extension Contexts (are there any? If so, how many?) OR if it's not easily detectable which JSON-LD extensions are being used (is it possible to detect extensions and map them to JSON-LD Contexts?).

Popular implementations do include their extension contexts, but there's a long tail of less popular ones, and some of them may have incorrect contexts (because incorrect context usually doesn't result in processing error on the receiving side).

>Is there a registry of some kind for Fediverse extensions outside of FEPs?

No, but there are two proposals attempting to tackle the problem:

- [FEP-cb76: Content Addressed Vocabulary](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-cb76.md) | [discussion](https://socialhub.activitypub.rocks/t/fep-cb76-content-addressed-vocabulary/2735)
- [FEP-2e40: The FEP Vocabulary Extension Process](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-2e40.md) | [discussion](https://socialhub.activitypub.rocks/t/fep-1570-the-fep-ontology-process/2972)

>Of those two options, I expect developers to footgun themselves far more often with option 2 than option 1... adding jcs to the name of the cryptosuite feels like an argument for option 2.

Looks like security-usability dilemma. I don't know how to resolve this, but here's how I think FEP-8b32 and FEP-c390 will be implemented in practice:

- [FEP-8b32: Object Integrity Proofs](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-8b32.md) describes how to sign activities. It would make sense for developers to agree on one cryptosuite and use it. Currently proposal recommends to use `JCS` + `SHA-256` + `RSASSA-PKCS1-v1_5`. This is the path of least resistance: RSA and SHA-256 are already used by implementations to generate HTTP signatures, so every actor has an RSA key. We can call this cryptosuite `fediverse-2023`.
- [FEP-c390: Identity Proofs](https://codeberg.org/fediverse/fep/src/branch/main/feps/fep-c390.md) describes how to link external identity to ActivityPub actor. There are many types of external identities: raw keys, PGP keys, blockchain wallets, various SSI systems. We can expect that people will experiment a lot, and if they don't find an appropriate cryptosuite, they will start inventing them. In that case proliferation of cryptosuites like `jcs-ecdsa-sha2-256-p256-mysuperapp-2023` seems inevitable.

index.html
msporny: This change has been applied in af98f8225f90b4fc2126ec7f7828741379622f10 along with an issue marker noting that the cryptosuite name is disputed and is currently being debated.

msporny:

PR #28, opened 2023-03-03
Additional Test vectors and Separation of Files

Wind4Greg:
Based on comments from @OR13 I updated to use the [ReSpec data-include](https://github.com/w3c/respec/wiki/data-include-format) mechanism to bring in test vector information rather than embedding this information in-line in the HTML document.

I added test cases for all currently planned types/cryptosuites (including JCS). Files for test vectors have be put in separate directories based on type/cryptosuite. I used refined text from @TallTed as a template for all cases. The test vector generation code has been put in the public repo [Wind4Greg/EdDSA-Test-Vectors](https://github.com/Wind4Greg/EdDSA-Test-Vectors).

Hopefully this will make it easier to update test vectors in cases of changes to example document content, cryptosuite naming, etc...

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/di-eddsa-2020/pull/28.html" title="Last updated on Mar 3, 2023, 5:35 PM UTC (3fc9ad7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/28/5a3b66a...Wind4Greg:3fc9ad7.html" title="Last updated on Mar 3, 2023, 5:35 PM UTC (3fc9ad7)">Diff</a>


TestVectors/DataInt/canonDocDataInt.txt
```diff
@@ -0,0 +1,8 @@
+<did:example:ebfeb1f712ebc6f1c276e12ec21> <https://www.w3.org/ns/credentials/issuer-dependent#degree> _:c14n0 .
+<http://example.edu/credentials/3732> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://example.org/examples#UniversityDegreeCredential> .
+<http://example.edu/credentials/3732> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> .
+<http://example.edu/credentials/3732> <https://www.w3.org/2018/credentials#credentialSubject> <did:example:ebfeb1f712ebc6f1c276e12ec21> .
+<http://example.edu/credentials/3732> <https://www.w3.org/2018/credentials#issuer> <https://example.edu/issuers/565049> .
+<http://example.edu/credentials/3732> <https://www.w3.org/ns/credentials/issuer-dependent#issuanceDate> "2010-01-01T00:00:00Z" .
```
msporny, line 6: This is effectively an undefined property, which we shouldn't be doing in our test examples. 

`issuanceDate` is no longer valid as of v2.0... we need to use `validFrom` and `validUntil` now for validity date ranges.

Let's settle on a better test document to start from. My expectation is that most of these files are auto-generated?

msporny:
Oof, crap. I wish I had seen this PR before merging the other one. Conflicts everywhere now... sorry about that.

Can you re-base this on top of `main`?

TestVectors/DataInt/canonDocDataInt.txt
```diff
@@ -0,0 +1,8 @@
...
+_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/issuer-dependent#BachelorDegree> .
```
msporny, line 7: Any time you see `https://www.w3.org/ns/credentials/issuer-dependent#` -- the term is undefined and we should do better in our examples so we don't introduce stuff that might confuse implementers.

index.html
```diff
@@ -812,152 +812,200 @@ <h2>Privacy Considerations</h2>
...
-The following test vectors are provided to assist with implementers.
+        For all representations the signer needs to generate a private/public key pair with the private key used 
+        for signing and the public key made available for verification. The 
+        [[MULTIBASE]]/[[MULTICODEC]] representation for the public key, <code>ed25519-pub</code>,
+        and the representation for the private key, <code>ed25519-priv</code>, are shown below.
       </p>
+      <pre class="example" title="Private and Public keys for Signature"
+          data-include="TestVectors/keyPair.json" data-include-format="text"></pre>
+      
+      <section id="VectDataIntegrity">
```
msporny, line 826: No need to specify `id` values most of the time, Respec generates them automatically.

If you do hand-code an `id` value, use all lowercase letters, numbers, and dashes only (or we might fail publication rules checking at W3C ... specifically the anchoring checks when we go to publish to W3C). Let the tools do the heavy lifting for us and auto-generate `id`.

msporny (changes requested):
Overall, this is a good direction. Noting the following things:

1. PR Preview is now broken since it doesn't know how to work well with `data-include`. https://pr-preview.s3.amazonaws.com/Wind4Greg/di-eddsa-2020/pull/28.html#VectDataIntegrity -- so, we either have hard to read escaped code in the spec, OR, we have a broken PR Preview. We can probably fix PR Preview once this PR is merged by referencing github to pull in the examples from the public Web (once they have a published file via Github pages). So, let's keep going in this direction for now.
2. The test vector directories should be named after the cryptosuites... so, `eddsa-2022`, `Ed25519Signature2020`, and `json-eddsa-2022`.
3. We need to use an easier to understand example for the test vectors. I've raised a PR in your test vectors repo to update the example used for all the test vectors: https://github.com/Wind4Greg/EdDSA-Test-Vectors/pull/1/files

I tried sticking w/ the base `v2` context and quickly found out that it's unworkable.

TestVectors/DataInt/canonDocDataInt.txt
msporny: Proposal for new test vectors here: https://github.com/Wind4Greg/EdDSA-Test-Vectors/pull/1/files

TestVectors/DataInt/canonDocDataInt.txt
msporny: New proposed test vectors base file here: https://github.com/Wind4Greg/EdDSA-Test-Vectors/pull/1/files

index.html
Wind4Greg: Will do. I put the `id` value in to help navigate in my IDE (VS code) otherwise it gets hard to find things in the raw HTML. If you have other tricks let me know.

TestVectors/DataInt/canonDocDataInt.txt
Wind4Greg: Accepted the new test vectors! I guess I shouldn't have left the old pull request when it overlapped with the new one. Will figure out the `rebase` stuff...

PR #29, opened 2023-03-05
Update test vectors based on new example document

Wind4Greg:
Due to making a mess of my forked repo had to start this pull request again. This PR separates out the test vector data into files separate from the base HTML document using ReSpec's `include-data` functionality. In addition, this addresses the following issues raised by @msporny:

1. No need to specify id values most of the time, Respec generates them automatically. If you do hand-code an id value, use all lowercase letters, numbers, and dashes only (or we might fail publication rules checking at W3C ... specifically the anchoring checks when we go to publish to W3C). Let the tools do the heavy lifting for us and auto-generate id. ==> I used your recommended convention in the few places where I used hand coded `id`. I used this so I could find my way around the HTML in my IDE (VS Code).
2. The test vector directories should be named after the cryptosuites... so, eddsa-2022, Ed25519Signature2020, and json-eddsa-2022. ==> Done.
3. We need to use an easier to understand example for the test vectors. I've raised a PR in your test vectors repo to update the example used for all the test vectors: https://github.com/Wind4Greg/EdDSA-Test-Vectors/pull/1/files ==> Thanks for the better example! Accepted pull request and re-generated test vector files based on you example document.

Note: The code for generating all the test vectors is available at [Wind4Greg/EdDSA-Test-Vectors](https://github.com/Wind4Greg/EdDSA-Test-Vectors).

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-di-eddsa/pull/29.html" title="Last updated on May 16, 2023, 4:04 PM UTC (b488dd0)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/29/3c8838d...Wind4Greg:b488dd0.html" title="Last updated on May 16, 2023, 4:04 PM UTC (b488dd0)">Diff</a>


TestVectors/Ed25519Signature2020/signedEdSig.json
```diff
@@ -0,0 +1,26 @@
...
+  "issuer": "https://vc.example/issuers/5678",
+  "validFrom": "2023-01-01T00:00:00Z",
+  "credentialSubject": {
+    "id": "did:example:abcdefgh",
+    "alumniOf": "The School of Examples"
+  },
+  "proof": {
+    "type": "Ed25519Signature2020",
+    "created": "2023-02-24T23:36:38Z",
+    "verificationMethod": "https://example.edu/issuers/565049#z6MkrJVnaZkeFzdQyMZu1cgjg7k1pZZ6pvBQ7XJPt4swbTQ2",
```
msporny:
```suggestion
    "verificationMethod": "https://vc.example/issuers/5678#z6MkrJVnaZkeFzdQyMZu1cgjg7k1pZZ6pvBQ7XJPt4swbTQ2",
```

We'll want to make this change across the board in all examples.

msporny:
Given that this PR reformats an earlier PR that had 3 weeks of circulation, multiple reviews, no objections, and a merge... and that this PR commits what should've gone in the other PR (reformatting requested by @OR13)... I'm going to merge this as soon as we have another implementer confirm that these test vectors look better than the ones we have in the spec right now. That might happen sooner than by next weekend, just giving reviewers a heads-up.

TestVectors/Ed25519Signature2020/signedEdSig.json
Wind4Greg: Updated proof option `VerificationMethod` across all examples and regenerated test vectors. Committed and pushed.

msporny (approved):

brentzundel:
@iherman not sure if there are any IPR concerns here. I do not believe @Wind4Greg is a VCWG participant.

Wind4Greg:
Hi all, I'm a CCG member. So there should not be IPR concerns, i.e., I "signed" the paper work.

msporny:
> Hi all, I'm a CCG member. So there should not be IPR concerns, i.e., I "signed" the paper work.

Unfortunately, it doesn't work like that. CCG IPR commitments do not apply to PRs raised in the VCWG. :)

This is a work item for the W3C Verifiable Credentials Working Group, which is a W3C Member-only group. You can still contribute to the group, but you MUST assign joint copyright over to W3C (so W3C can eventually publish it as a patent and royalty-free global standard) via an "Invited Expert" agreement and at least one of the following things needs to hold for all of your PRs:

* It does not contain anything that could be in a patent that you or others have, 
* if it contains any IP (at a minimu, your copyright), you release it for use in the standard in a patent and royalty-free manner.

Given that these are just test vectors that test the specification, I don't expect there's much concern here... but if you plan to make more PRs (which I think you do), we probably need to get you into the group as an Invited Expert.

/cc @iherman @Sakurann @brentzundel

Wind4Greg:
Ah, CCG versus VCWG... Happy to do whatever needed on the IPR front.

dlongley:
So I was able to make our implementation of `eddsa-2022` match the test vector here -- but since the `examples/v2` context hasn't been published to the URL specified, I had some trouble finding the right context to use. I used the `examples/v1` context and replaced the term definitions to get them to match with what was used here and then the signatures matched. This is just a temporary issue though, that will sort itself out once the contexts stabilize -- we might just need to regen test vectors as a result.

There was one other difference that we will want to sort out. My understanding is that we wanted to use `secretKeyMultibase` over `privateKeyMultibase` (the spec says the latter today) -- and that we should probably use the 64 byte key representation instead of the 32 byte key representation. The 64 byte key representation just concatenates the public key value after the private key value, which is what most implementations, I believe, use.

dlongley (approved):
Based on my comments above, I think we should accept these and sort out any differences later. If we're able to make the key change before this goes in that's better, but not a hard requirement.

dlongley:
Another difference is with the JCS version and whether the `@context` is added to the proof options. We'll want to be clear about what to do there -- I'd still expect it to be present. The test vectors do not add it -- and if our implementation removes it, then test vector values match.

TestVectors/Ed25519Signature2020/proofCanonEdSig.txt
```diff
@@ -0,0 +1,4 @@
+_:c14n0 <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/issuer-dependent#Ed25519Signature2020> .
```
dlongley: This is incorrect due to the missing `https://w3id.org/security/suites/ed25519-2020/v1` context.

dlongley (changes requested):
Actually, there's an important difference that I found with the `Ed25519Signature2020` representation. It's missing the `https://w3id.org/security/suites/ed25519-2020/v1` context. We should probably fix this before merging as it's incorrect and will change the test vector results.

TestVectors/Ed25519Signature2020/sigBTC58EdSig.txt
```diff
@@ -0,0 +1 @@
+z2fa386qEULLhQvuusYwE1wfjb7i6p9Ba15m83CpZne9ZgsiZgX8bUV3Ucdx8atbUgyskBkeouMHA9ay2G8Yqeu6n
```
dlongley: If nothing else needs to change after the missing context is added, this is the expected value:

```suggestion
zveu47DychpPUMPjgyDdFynHQSrXWZLy75uPLn8TB6aSyFRRvJHW7BHL4Nkyg1sSQWQZTa5196X8jo7EZMWEjKuz
```

TestVectors/Ed25519Signature2020/proofCanonEdSig.txt
Wind4Greg: I'm not quite sure about all the @context fields currently in the draft. When I look at Examples 5 and 6 they just use "https://w3id.org/security/data-integrity/v1" and not "https://w3id.org/security/suites/ed25519-2020/v1". "https://w3id.org/security/data-integrity/v1" was subsumed into "https://www.w3.org/ns/credentials/v2".

The ed25519-2020 context contains some of the same fields. Shouldn't the draft tell us which contexts to use in each case?

Wind4Greg: For the JCS version, there wasn't any text to go by. Since it is an extra step to put the `@context` back into the proof options and the `@context` isn't needed for JCS (JSON canonicalization scheme). I left out the step.

Wind4Greg: Concerning: "There was one other difference that we will want to sort out. My understanding is that we wanted to use secretKeyMultibase over privateKeyMultibase (the spec says the latter today) -- and that we should probably use the 64 byte key representation instead of the 32 byte key representation. The 64 byte key representation just concatenates the public key value after the private key value, which is what most implementations, I believe, use."

I think @msporny raised a point like this before. I'm not sure that there will be a standard for an unwrapped private/secret key though [multicodec](https://ipfs.io/ipfs/QmXec1jjwzxWJoNbxQF5KffL8q6hFXm9QwUGaa3wKGk6dT/#title=Multicodecs&src=https://raw.githubusercontent.com/multiformats/multicodec/master/table.csv) lists some. Most documents with test vectors that include secret key just give the HEX. Maybe we should just do that.

I would find it really confusing to call something a private/secret key and then find its the wrong length because it has a public key appended.

dlongley: @Wind4Greg,

> I would find it really confusing to call something a private/secret key and then find its the wrong length because it has a public key appended.

Yes, that may be true, but I think that's just a consequence of a bigger problem: having key expressions that fail to encapsulate crypto primitive internal components and failing to have application layer code treat crypto primitive expressions as opaque blobs.

Some key expressions, for example some key types expressed as JWKs, suffer from this -- whereby individual components of cryptographic keys are delimited and exposed in the JSON expression of the key, rather than encapsulating everything in a blob that is handed to the crypto-layer implementation. I think this is a layering violation and a mistake that can result in a need for more validation code and other bugs / confusion / attack surface.

Ultimately, at this layer, we shouldn't have to care what's in the field -- just what the field means ("it's a 'secret key of type X'") and how the value is base-encoded. That's all we should need to decode the thing and pass it to the right crypto-layer implementation. Whether crypto implementations standardize on private key expressions that include the public key concatenated onto the end or not should be totally irrelevant to us at this layer -- we should treat these things as opaque and just pass them through.

Once we find ourselves validating whether some point is compressed to just the `x` coordinate and it can expand out the `y` coordinate or if the Chinese Remainder theorem components are present and work properly with one another -- we've failed at this layer, IMO.

So, at this layer, I think we should not care what the blob of bytes is for the private key -- we just know that's what it is. That blob should conform to what most implementations accept. Now, for whatever reason, that happens to be a 64 byte array with the private key seed + associated public key concatenated onto it. That this was the choice that most crypto-layer implementations made shouldn't matter to us, we should just say that's how it is and run with it, IMO.

Wind4Greg: I think we are in agreement! I'm still not sure about what should be in the `@context` fields of examples (5 and 6) in the draft and the test vector unsigned document. Will we really need two different test vector unsigned base documents just differing in their `@context` fields? Maybe since one maybe for legacy. I'd like @msporny to chime in since he suggested the base test vector document. I can rerun test vectors relatively easily.

TestVectors/Ed25519Signature2020/canonDocEdSig.txt
```diff
@@ -0,0 +1,8 @@
+<did:example:abcdefgh> <https://schema.org/alumniOf> "The School of Examples" .
+<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> .
+<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/examples#AlumniCredential> .
+<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <https://schema.org/description> "An minimum viable example of an Alumni Credential." .
```
TallTed:
```suggestion
<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <https://schema.org/description> "A minimum viable example of an Alumni Credential." .
```

TestVectors/Ed25519Signature2020/signedEdSig.json
```diff
@@ -0,0 +1,26 @@
...
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33",
+  "type": [
+    "VerifiableCredential",
+    "AlumniCredential"
+  ],
+  "name": "Alumni Credential",
+  "description": "An minimum viable example of an Alumni Credential.",
```
TallTed:
```suggestion
  "description": "A minimum viable example of an Alumni Credential.",
```

TestVectors/eddsa-2022/canonDocDataInt.txt
```diff
@@ -0,0 +1,8 @@
+<did:example:abcdefgh> <https://schema.org/alumniOf> "The School of Examples" .
+<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/2018/credentials#VerifiableCredential> .
+<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <https://www.w3.org/ns/credentials/examples#AlumniCredential> .
+<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <https://schema.org/description> "An minimum viable example of an Alumni Credential." .
```
TallTed:
```suggestion
<urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33> <https://schema.org/description> "A minimum viable example of an Alumni Credential." .
```

TestVectors/eddsa-2022/signedDataInt.json
```diff
@@ -0,0 +1,27 @@
...
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33",
+  "type": [
+    "VerifiableCredential",
+    "AlumniCredential"
+  ],
+  "name": "Alumni Credential",
+  "description": "An minimum viable example of an Alumni Credential.",
```
TallTed:
```suggestion
  "description": "A minimum viable example of an Alumni Credential.",
```

TestVectors/json-eddsa-2022/canonDocJCS.txt
```diff
@@ -0,0 +1 @@
+{"@context":["https://www.w3.org/ns/credentials/v2","https://www.w3.org/ns/credentials/examples/v2"],"credentialSubject":{"alumniOf":"The School of Examples","id":"did:example:abcdefgh"},"description":"An minimum viable example of an Alumni Credential.","id":"urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33","issuer":"https://vc.example/issuers/5678","name":"Alumni Credential","type":["VerifiableCredential","AlumniCredential"],"validFrom":"2023-01-01T00:00:00Z"}
```
TallTed:
```suggestion
{"@context":["https://www.w3.org/ns/credentials/v2","https://www.w3.org/ns/credentials/examples/v2"],"credentialSubject":{"alumniOf":"The School of Examples","id":"did:example:abcdefgh"},"description":"A minimum viable example of an Alumni Credential.","id":"urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33","issuer":"https://vc.example/issuers/5678","name":"Alumni Credential","type":["VerifiableCredential","AlumniCredential"],"validFrom":"2023-01-01T00:00:00Z"}
```

TestVectors/json-eddsa-2022/signedJCS.json
```diff
@@ -0,0 +1,27 @@
...
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
+  ],
+  "id": "urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33",
+  "type": [
+    "VerifiableCredential",
+    "AlumniCredential"
+  ],
+  "name": "Alumni Credential",
+  "description": "An minimum viable example of an Alumni Credential.",
```
TallTed:
```suggestion
  "description": "A minimum viable example of an Alumni Credential.",
```

TestVectors/unsigned.json
```diff
@@ -0,0 +1,16 @@
+{
+    "@context": [
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
+    ],
+    "id": "urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33",
+    "type": ["VerifiableCredential", "AlumniCredential"],
+    "name": "Alumni Credential",
+    "description": "An minimum viable example of an Alumni Credential.",
```
TallTed:
```suggestion
    "description": "A minimum viable example of an Alumni Credential.",
```

TallTed (changes requested):
I think these (tiny) changes will require recalculations of all related crypto...

TestVectors/unsigned.json
Wind4Greg: Good catch @TallTed I'll fix the base unsigned document to be something like:

```json
{
    "@context": [
        "https://www.w3.org/ns/credentials/v2",
        "https://www.w3.org/ns/credentials/examples/v2"
    ],
    "id": "urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33",
    "type": ["VerifiableCredential", "AlumniCredential"],
    "name": "Alumni Credential",
    "description": "A minimum viable example of an Alumni Credential.",
    "issuer": "https://vc.example/issuers/5678",
    "validFrom": "2023-01-01T00:00:00Z",
    "credentialSubject": {
        "id": "did:example:abcdefgh",
        "alumniOf": "The School of Examples"
    }
}
```

I say something like this, because I'm still trying to resolve with the JSON-LD experts what should be the `@context` and whether I need two or more different unsigned documents for the test vector generation.

TestVectors/Ed25519Signature2020/canonDocEdSig.txt
OR13: ^ probably not a good idea to change the n-quads, this change should be made to the JSON-LD.

TestVectors/Ed25519Signature2020/proofConfigEdSig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "Ed25519Signature2020",
+  "created": "2023-02-24T23:36:38Z",
+  "verificationMethod": "https://vc.example/issuers/5678#z6MkrJVnaZkeFzdQyMZu1cgjg7k1pZZ6pvBQ7XJPt4swbTQ2",
+  "proofPurpose": "assertionMethod",
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 8:
```suggestion
```

OR13 (changes requested):
proof contains invalid context values.

TestVectors/Ed25519Signature2020/signedEdSig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

OR13: As you can see from the n-quads, this value is not required to generate test vectors for this suite.

TestVectors/Ed25519Signature2020/signedEdSig.json
OR13: This change should be accepted.

TestVectors/eddsa-2022/proofConfigDataInt.json
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "DataIntegrityProof",
+  "cryptosuite": "eddsa-2022",
+  "created": "2023-02-24T23:36:38Z",
+  "verificationMethod": "https://vc.example/issuers/5678#z6MkrJVnaZkeFzdQyMZu1cgjg7k1pZZ6pvBQ7XJPt4swbTQ2",
+  "proofPurpose": "assertionMethod",
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 9:
```suggestion
```

OR13: Again, not needed... this context is 404, and it defines no terms in the proof.

TestVectors/eddsa-2022/signedDataInt.json
```diff
@@ -0,0 +1,27 @@
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

TestVectors/json-eddsa-2022/canonDocJCS.txt
OR13:
```suggestion
{"@context":["https://www.w3.org/ns/credentials/v2"],"credentialSubject":{"alumniOf":"The School of Examples","id":"did:example:abcdefgh"},"description":"An minimum viable example of an Alumni Credential.","id":"urn:uuid:58172aac-d8ba-11ed-83dd-0b3aef56cc33","issuer":"https://vc.example/issuers/5678","name":"Alumni Credential","type":["VerifiableCredential","AlumniCredential"],"validFrom":"2023-01-01T00:00:00Z"}
```

TestVectors/json-eddsa-2022/signedJCS.json
```diff
@@ -0,0 +1,27 @@
+{
+  "@context": [
+    "https://www.w3.org/ns/credentials/v2",
+    "https://www.w3.org/ns/credentials/examples/v2"
```
OR13:
```suggestion
```

TestVectors/unsigned.json
```diff
@@ -0,0 +1,16 @@
+{
+    "@context": [
+        "https://www.w3.org/ns/credentials/v2",
+        "https://www.w3.org/ns/credentials/examples/v2"
```
OR13, line 4:
```suggestion
```

OR13:
Related PR: https://github.com/w3c/vc-data-model/pull/1066

TestVectors/eddsa-2022/proofConfigDataInt.json
Wind4Greg: Two separate issues here. The context being 404 and the not a `@context` value defining terms not in the proof configuration.

When I produced this set of test vectors both contexts "https://www.w3.org/ns/credentials/examples/v2", and "https://www.w3.org/ns/credentials/v2" were undefined. Now only the first is undefined. These `@context` values were recommended by @msporny as being forward looking and up to date (too up to date?). I'm assuming the first one will be updated/published soon.

Regarding the `@context` for `proofConfig` the current specification leaves out the canonicalization step for `proofConfig`. To do this step ("Universal RDF Dataset Canonicalization") `@context` needs to be applied to the generated `proofConfig`. I took a robust approach of copying the `@context` from the unsigned document to the `proofConfig` during test vector generation.

However, in the JCS case I did not do this since it seemed unnecessary (JCS canonicalization doesn't use context), but this raised issues with those test vectors from @dlongley for not adding the `@context` to proof config.

Hence I think we need to refine the specification on `proofConfig` generation and canonicalization.

This does show the power of test vectors in helping move the specification along! Cheers Greg

TestVectors/Ed25519Signature2020/proofCanonEdSig.txt
dlongley: Additional good rationale for the secret key to be the 64-byte concatenation of the secret seed and public key point: https://crypto.stackexchange.com/questions/54353/why-are-nacl-secret-keys-64-bytes-for-signing-but-32-bytes-for-box

So I think we should make clear that that is what it is in our specs.

Wind4Greg: Sounds good.

TestVectors/eddsa-2022/proofConfigDataInt.json
iherman: > When I produced this set of test vectors both contexts "https://www.w3.org/ns/credentials/examples/v2", and "https://www.w3.org/ns/credentials/v2" were undefined. Now only the first is undefined. These @context values were recommended by @msporny as being forward looking and up to date (too up to date?). I'm assuming the first one will be updated/published soon.

@Wind4Greg https://www.w3.org/ns/credentials/v2 is now live and points to the most up-to-date JSON-LD context for the VCDM. It also includes an extra `@vocab` to https://www.w3.org/ns/credentials/issuer-dependent# which is also live and replaces the usage of an extra context file at https://www.w3.org/ns/credentials/examples/v2 (which does not, and will not, be valid and will stay 404). Ie, the references to the latter should be removed from all the examples.

msporny: > replaces the usage of an extra context file at https://www.w3.org/ns/credentials/examples/v2 (which does not, and will not, be valid and will stay 404). Ie, the references to the latter should be removed from all the examples.

Wait, no, big -1 on not defining a v2 examples context.

I am suggesting that we do the following in the spec:

1. Have a section that outlines "How to get started", that shows examples w/ no secondary context.
2. Clearly articulate that the ideal case is that people define clear semantics in their VCs via a secondary context and use that... and then the rest of the specification uses real world examples that use a secondary context, or the examples context (where we need a carefully crafted example).

OR13: You could define a v2 examples context and add it to the vc-specs dir.
You could also define a "data integrity examples" context.

If the VCWG is going to define a v2 examples context again, it must not load other contexts, and it should be minimized, and ideally not biased towards one vertical (ed tech).

iherman: > I am suggesting that we do the following in the spec:
> 
> 1. Have a section that outlines "How to get started", that shows examples w/ no secondary context.
> 2. Clearly articulate that the ideal case is that people define clear semantics in their VCs via a secondary context and use that... and then the rest of the specification uses real world examples that use a secondary context, or the examples context (where we need a carefully crafted example).

I am not against that but, if we do, we should have _something_ at the receiving end of the URL. I am not saying we should have a formal vocabulary, but at least a landing page, akin to what we now at the `.../issuer-dependent` URL.

OR13: ahh, yes... totally agree... maybe  `@vocab`: `https://example.edu/vocab#` if the examples are exclusive to education tech.

iherman:
> @iherman not sure if there are any IPR concerns here. I do not believe @Wind4Greg is a VCWG participant.

There is no IPR concern here (it is only on testing), so the PR is cleared for merge (in this respect).

index.html
```diff
@@ -1184,6 +1184,152 @@ <h4>Proof Verification (Ed25519Signature2020)</h4>
...
+      data-include="TestVectors/eddsa-2022/proofConfigDataInt.json" data-include-format="text"></pre>
+
+      <pre class="example" title="Canonical Proof Options Document"
+      data-include="TestVectors/eddsa-2022/proofCanonDataInt.txt" data-include-format="text"></pre>
+
+      <pre class="example" title="Hash of Canonical Proof Options Document (hex)"
+      data-include="TestVectors/eddsa-2022/proofHashDataInt.txt" data-include-format="text"></pre>
+
+        <p>
+Finally, we combine the two hashes, use the private key with the combined hash to
```
TallTed: How are the hashes combined? Is it a simple concatenation? Should say whatever it is... 

Also, if important (as it seems it is), the order of the hashes in the combination should be specified.

(These answers may be obvious in the included text files, but [they're not visible in the PR-preview](https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-di-eddsa/pull/29.html#example-combine-hashes-of-proof-options-and-credential-hex) nor do they appear to be [within this repo](https://github.com/w3c/vc-di-eddsa/search?q=combinedHashEdSig.txt).)

index.html
```diff
@@ -1184,6 +1184,152 @@ <h4>Proof Verification (Ed25519Signature2020)</h4>
...
+compute the Ed25519 signature, and then base58-btc encode the signature.
+        </p>
+
+      <pre class="example" title="Combine hashes of Proof Options and Credential (hex)"
+      data-include="TestVectors/eddsa-2022/combinedHashDataInt.txt" data-include-format="text"></pre>
+
+      <pre class="example" title="Signature of Combined Hashes (hex)"
+      data-include="TestVectors/eddsa-2022/sigHexDataInt.txt" data-include-format="text"></pre>
+
+      <pre class="example" title="Signature of Combined Hashes base58-btc"
```
```suggestion
      <pre class="example" title="Signature of Combined Hashes (base58-btc)"
```

index.html
```diff
@@ -1184,6 +1184,152 @@ <h4>Proof Verification (Ed25519Signature2020)</h4>
...
+compute the Ed25519 signature, and then base58-btc encode the signature.
+      </p>
+
+    <pre class="example" title="Combine hashes of Proof Options and Credential (hex)"
+    data-include="TestVectors/json-eddsa-2022/combinedHashJCS.txt" data-include-format="text"></pre>
+
+    <pre class="example" title="Signature of Combined Hashes (hex)"
+    data-include="TestVectors/json-eddsa-2022/sigHexJCS.txt" data-include-format="text"></pre>
+
+    <pre class="example" title="Signature of Combined Hashes base58-btc"
```
```suggestion
    <pre class="example" title="Signature of Combined Hashes (base58-btc)"
```

TallTed (changes requested):

index.html
```diff
@@ -29,8 +29,8 @@
         // previousMaturity:  "WD",
 
         // if there a publicly available Editor's Draft, this is the link
-        edDraftURI: "https://w3c.github.io/vc-di-eddsa/",
-        //latestVersion: "https://www.w3.org/community/reports/credentials/CG-FINAL-di-eddsa-2020-20220724/",
+        edDraftURI: "https://w3c-ccg.github.io/di-eddsa-2020/",
+        latestVersion: "https://www.w3.org/community/reports/credentials/CG-FINAL-di-eddsa-2020-20220724/",
```
davidlehn: Looks like this change should be reverted.

index.html
```diff
@@ -65,7 +65,21 @@
...
+        wgURI: "https://www.w3.org/community/credentials/",
+        // name (with the @w3c.org) of the public mailing to which comments are due
+        wgPublicList: "public-credentials",
+        // URI of the patent status for this WG, for Rec-track documents
+        // !!!! IMPORTANT !!!!
+        // This is important for Rec-track documents, do not copy a patent URI from a random
+        // document unless you know what you're doing. If in doubt ask your friendly neighbourhood
+        // Team Contact.
+        wgPatentURI: "https://www.w3.org/community/about/agreements/cla/",
+        github: "https://github.com/w3c-ccg/di-eddsa-2020/",
```
davidlehn: Looks like this change should be reverted.  Check other changes to see if they slipped in by mistake during a merge.

index.html
```diff
@@ -433,10 +442,10 @@ <h4>Transformation (eddsa-2022)</h4>
 identifier (<var>cryptosuite</var>). A <em>transformed data document</em> is
 produced as output. Whenever this algorithm encodes strings, it MUST use UTF-8
 encoding.
-          </p>
+        </p>
```
davidlehn: Looks like these indent changes are reverting the fixes from https://github.com/w3c/vc-di-eddsa/pull/34.  Can you please undo all of them as appropriate?  Sorry for the conflict troubles!

index.html
```diff
@@ -1292,38 +1254,156 @@ <h3>Representation: Ed25519Signature2020</h3>
           </li>
         </ol>
 
-        <pre class="example" title="Signed Credential">
+      <pre class="example" title="Signed Credential"
+      data-include="TestVectors/eddsa-2022/signedDataInt.json" data-include-format="text"></pre>
+    </section>
+    <section>
```
davidlehn: Related to indentation issue above, most tags in this section need to go right 2 spaces.  Though some pre/p tags are mixed up levels.  They should all end up at properly tag tree levels in the end.

index.html
Wind4Greg: How much should we repeat the algorithm steps given in the normative text? These should match the steps of section 3. Can add more text. Kind of stuck waiting on agreement on `@context` field for base document.

Cheers Greg

index.html
Wind4Greg: Looks like things got screwed up when I tried to merge w3c repo `main` into my main. Probably need to start again...

index.html
Wind4Greg: Any easy way to do that?

index.html
Wind4Greg: Any easy way to do that?

index.html
Wind4Greg: Any recommendations for tools to automate this stuff? Hand indentation is quite tedious. I usually just use various auto formatters for my code.

davidlehn: I have vim setup in some way that I go top-down to lines with tags that are off and hit '==' and it fixes it.  And I just go through until done.  A bit tedious, but it doesn't take long, usually.  Probably takes as long to write and revise this comment. :-)  I'm not sure what other tools would fix this without causing mass havoc on all the other markup.  With the style in many of these specs you can also grep for tags and see the indentation that way to visually find issues.

index.html
TallTed: A link from each shortened step description here, back to the detailed step description in the normative text (which details would presumably answer my questions here), would help people who skip ahead to the examples (as many will do, no matter the extent of harm they'd potentially be doing themselves).

As to these particular questions, it seems to me that saying "we concatenate hash x and hash y in that order" would be substantially clearer than "we combine the two hashes" without adding much text — and linking back to the normative text would still be beneficial.

index.html
Wind4Greg: We used Vim in the '80s at Berkeley. I'll have to try something that works with VS Code...

index.html
Wind4Greg: Changes should be reverted. Started with w3c main then added back in test vectors, and hand crafting indentation.

TestVectors/Ed25519Signature2020/canonDocEdSig.txt
TallTed: OK, so this is an artifact, not a managed document, and should probably be added to `.gitignore`.

TestVectors/eddsa-2022/canonDocDataInt.txt
TallTed: Another artifact?

TestVectors/eddsa-2022/signedDataInt.json
TallTed: Another artifact?

TestVectors/json-eddsa-2022/canonDocJCS.txt
TallTed: Another artifact?

TestVectors/json-eddsa-2022/signedJCS.json
TallTed: Another artifact?

TestVectors/Ed25519Signature2020/canonDocEdSig.txt
Wind4Greg: No, it is a referenced (included document), so cannot be added to .gitignore. When we update the base test vector document it will be re-created but that does not happen in this repo.

TestVectors/eddsa-2022/canonDocDataInt.txt
Wind4Greg: Artifact? Quite a while ago it was requested that all test vector be handled via ReSpec's `data-include` mechanism rather than embedded/escaping lots of JSON and other information directly into the `index.html` file. These are kept in a separate `TestVectors` directory and included via ReSpec mechanisms.

TallTed: Either I'm commenting on artifacts — indirectly managed files produced from directly managed files by CI or similar processes — in which case these indirectly managed files should be added to `.gitignore` or similar so they don't show up in the `/files` view of a given PR, *OR* I'm commenting on those directly managed files, in which case I request that my suggestions be addressed with somewhat more speed.

I believe my suggestions should generally be accepted/merged into the PR fork, but you or others may disagree with my suggestions, in which case comments in reply to my suggestions should reflect such disagreement, so that we may work to resolve misunderstanding or other basis for the disagreement, and come to agreement on appropriate change (which may include retraction of my suggestion).

Wind4Greg:
## Including Test Vectors in W3C Specifications

This specification includes procedures for cryptographic processing of content. To aid developers in implementing specifications of this type. We provide test vectors illustrating this processing. These test vectors also demonstrate intermediate processing steps as well as initial inputs and outputs. This specification is written in HTML along with the [ReSpec](https://github.com/w3c/respec) library. The raw test vector data is text based, e.g., JSON, N-quads, etc... and generally requires significant "HTML escaping" if embedded directly into the document.

ReSpec has a [data-include](https://github.com/w3c/respec/wiki/data-include) directive that can be used to include separate files into a specification when it is viewed. Such files **must be kept** along with the `index.html` document for the specification. We have currently chosen to put the test vectors in the directory `TestVectors`. As an example, to show the unsigned document we use the following HTML/ReSpec:

```html
<pre class="example" title="Credential without Proof" data-include="TestVectors/unsigned.json" 
data-include-format="text"></pre>
```

ReSpec will pull in the content from the referenced file when the `index.html` document is viewed. Note that when viewing the `index.html` locally you will need to run a local webserver, e.g., [serve](https://www.npmjs.com/package/serve), since the ReSpec library needs a server to deliver the referenced files. Hence you cannot simply "double click" the `index.html` file to preview the document.

Note that the test vector files are generated by programs outside this repository that themselves depend on other libraries and programming languages, hence we probably don't want to include those programs in this repo and hence should just revision control the test vector files.

### What's Important to Review

Everything of course but some things stem from others...
In out case the two main inputs to test vector generation are `TestVectors/unsigned.json` and `TestVectors/keyPair.json`. The explanatory text is all in updates to `index.html` and a *informative* section towards the end of the document. Note we have three different processing cases which leads to three similar sets of test vectors and their descriptions. Hence consider if you want changes to `index.html` concerning the test vector to apply to all cases.

TestVectors/eddsa-2022/proofConfigDataInt.json
msporny: Ok, finally found these set of comments... I thought I had hallucinated them.

 I'm on board w/ what @OR13 and @iherman are suggesting above... once we have that in VCDM, we can use the v2 examples context (not bound to any market vertical... using something like: `"@vocab": "https://www.w3.org/credentials/examples#"` and then that points to a document, much like the issuer-dependent one, but clearly articulates that the vocabulary is only to be used for examples.

OR13: yep, I don't think W3C should be hosting example context files... This would be an acceptable solution:

```json
"@context": [
    "https://www.w3.org/ns/credentials/v2",
    "https://www.w3.org/ns/credentials/examples/v2"
```

changed to:

```json
"@context": [
    "https://www.w3.org/ns/credentials/v2",
    { "@vocab": "https://www.w3.org/credentials/examples#" }
```

And then all the terms would result in definitions that are just as "defined" as this:

```json
"@context": [
    "https://www.w3.org/ns/credentials/v2"
```

(In the sense that thee term IRIs don't dereference to anything useful and the URL communicates that is going to be the case).

Wind4Greg: Looks like @msporny and @OR13 are starting to converge? What would really be helpful is to agree on the contents of the unsigned document (example document) that is input to the process, i.e., `TestVectors/unsigned.json`.

Cheers Greg

dlongley: Of course, using `@vocab` at all here means that implementations that want to prohibit usage of `@vocab` will have to build special flags to enable its use when generating test vectors. That seems unfortunate, but doable.

msporny:
@Wind4Greg This JSON-LD Context URL now exists: https://www.w3.org/ns/credentials/examples/v2

That clears the way for this PR.

Please clear up whatever other merge requests need to be handled in this PR and let's get these test vectors into the spec.

TestVectors/Ed25519Signature2020/proofConfigEdSig.json
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TestVectors/Ed25519Signature2020/signedEdSig.json
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TestVectors/eddsa-2022/proofConfigDataInt.json
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TestVectors/eddsa-2022/signedDataInt.json
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TestVectors/json-eddsa-2022/canonDocJCS.txt
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TestVectors/json-eddsa-2022/signedJCS.json
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TestVectors/unsigned.json
msporny: https://www.w3.org/ns/credentials/examples/v2 exists now and thus this change request is no longer valid.

TallTed (changes requested):
* Every instance of `An minimum` should be changed to `A minimum`. I do not see that this has happened yet.

* The comments I made about "combine the two hashes" (circa line 1222) do not appear to have been addressed. I still believe this (and similar, e.g., circa line 1368) text should be clarified as I've already described circa line 1222, and at least one `<a href...>` or similar added to link back to the spec upon which this example is based.

Wind4Greg:
See latest changes to address:

1. @OR13 issues with `@context`
2. @TallTed issue with "A minimum viable..."
3. @TallTed issue with explanation of hash concatenation in test vector.

TestVectors/unsigned.json
TallTed: This instance of `An minimum` does not appear to have been fixed. All other instances appear to have been fixed.

TallTed:
Hash combination details look good.

One instance of `An minimum` appears to remain (also noted inline).

TallTed (approved):

OR13 (approved):
v2 example context now resolves, hopefully these test vectors are now accurate.

msporny:
Editorial (but substantive), multiple reviews, changes requested and made, no objections, merging.

PR #42, opened 2023-04-26
Update Sig2020 appendix for proof canonicalization.

Wind4Greg:
Appendix A "The Ed25519Signature2020 Suite" is missing the canonicalization step for the *proof configuration*. This PR add the relevant steps and update. These steps are essentially the same as used in section "3.1.5 Proof Configuration (eddsa-2022)" and are necessary to conform with existing implementations and test vectors

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-di-eddsa/pull/42.html" title="Last updated on Apr 26, 2023, 5:42 PM UTC (b696cb1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/42/14022e4...Wind4Greg:b696cb1.html" title="Last updated on Apr 26, 2023, 5:42 PM UTC (b696cb1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1016,7 +1016,7 @@ unlinkability. If signatures are re-used, they can be used as correlatable data.
               <li>
   Let <var>proofConfigHash</var> be the result of applying the
   SHA-256 (SHA-2 with 256-bit output) cryptographic hashing algorithm [[RFC6234]]
-  to the <var>proofConfig</var>. <var>proofConfigHash</var> will be
+  to the <var>canonicalProofConfig</var>. <var>proofConfigHash</var> will be
   exactly 32 bytes in size.
               </li>
   
@@ -1080,8 +1080,17 @@ unlinkability. If signatures are re-used, they can be used as correlatable data.
   <var>options</var>.<var>proofPurpose</var>.
               </li>
               <li>
-  Return <var>proofConfig</var>.
+  Set <var>proofConfig</var>.<var>@context</var> to
+  <var>unsecuredDocument</var>.<var>@context</var>
               </li>
+              <li>
+  Let <var>canonicalProofConfig</var> be the result of applying the
+  Universal RDF Dataset Canonicalization Algorithm
+  [[RDF-CANON]] to the <var>proofConfig</var>.
+              </li>
+              <li>
+  Return <var>canonicalProofConfig</var>.
+              </li>  
             </ol>
   
           </section>


iherman:
@Wind4Greg it is really great to see you contributing to these specs. The only thing is that these IPR warnings will keep popping up which is tiring for all of us I presume. Can you contact @brentzundel and @Sakurann to see whether you could sign up to the WG as an invited expert? That would make all this much smoother.

In the meantime... @msporny @dmitrizagidulin I would like your opinion whether these proposed changes are IPR sensitive or not. My gut feeling is they are not, but it is not for me to judge...

Thx

index.html
```diff
@@ -1080,8 +1080,17 @@ <h4>Proof Configuration (Ed25519Signature2020)</h4>
...
               </li>
               <li>
-  Return <var>proofConfig</var>.
+  Set <var>proofConfig</var>.<var>@context</var> to
+  <var>unsecuredDocument</var>.<var>@context</var>
               </li>
+              <li>
+  Let <var>canonicalProofConfig</var> be the result of applying the
+  Universal RDF Dataset Canonicalization Algorithm
+  [[RDF-CANON]] to the <var>proofConfig</var>.
```
msporny, line 1089:
```suggestion
  Let <var>canonicalProofConfig</var> be the result of applying the
  Universal RDF Dataset Canonicalization Algorithm
  [[RDF-CANON]] to the <var>proofConfig</var>.
```

Good catch! Note that a similar change will need to be made to the jcs-eddsa-2022 and ecdsa-2019 algorithms as well.

index.html
```diff
@@ -1016,7 +1016,7 @@ <h4>Hashing (Ed25519Signature2020)</h4>
               <li>
   Let <var>proofConfigHash</var> be the result of applying the
   SHA-256 (SHA-2 with 256-bit output) cryptographic hashing algorithm [[RFC6234]]
-  to the <var>proofConfig</var>. <var>proofConfigHash</var> will be
+  to the <var>canonicalProofConfig</var>. <var>proofConfigHash</var> will be
```
msporny, line 1019: This algorithm doesn't explain where `canonicalProofConfig` comes from... (I believe it comes in as input to the function, but we don't say that in the opening paragraph ... and probably should via a `<var>` declaration).

msporny (changes requested):
Overall +1, with the only requested change being to say where `canonicalProofConfig` comes from.

msporny:
> In the meantime... @msporny @dmitrizagidulin I would like your opinion whether these proposed changes are IPR sensitive or not. My gut feeling is they are not, but it is not for me to judge...

They are not IPR sensitive, we can clear the IPR flag on this item.

index.html
dlongley: Well, the "Proof Configuration" algorithm below now returns `canonicalProofConfig` *as* the `proofConfig`. So the only way to generate a `proofConfig` (which is an input to this algorithm here already) involves running that canonicalization step. So I don't think we need this change here -- and then we don't need to say where `canonicalProofConfig` comes from, because for `Ed25519Signature2020`, `proofConfig` *is* canonical / `canonicalProofConfig`.

So it seems that this PR already does what we need without any changes (without @msporny's suggestion here)?

dlongley (approved):

brentzundel:
@Wind4Greg has joined the VCWG as an invited expert.

dmitrizagidulin (approved):
👍

TallTed (commented):
lgtm, fwiw

msporny (approved):

msporny:

PR #51, opened 2023-07-24
Add EdDSA mitigation for dataset poisoning.

msporny:
This PR implements a new requirement to detect dataset poisoning by default in VC Data Integrity PR https://github.com/w3c/vc-data-integrity/pull/128

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/pull/51.html" title="Last updated on Aug 10, 2023, 1:12 AM UTC (61ca6d9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/51/792a747...61ca6d9.html" title="Last updated on Aug 10, 2023, 1:12 AM UTC (61ca6d9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -402,9 +402,9 @@ section also include the verification of such a data integrity proof.
         </p>
 
         <p class="advisement">
-When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is utilized,
+When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is used,
 implementations MUST detect <a data-cite="RDF-CANON#dataset-poisoning">
-dataset poisoning</a> by default and abort processing.
+dataset poisoning</a> by default, and abort processing upon such detection.
         </p>
 
         <section>
--- a/index.html
+++ b/index.html
@@ -401,6 +401,12 @@ resulting in the production of a data integrity proof. The algorithms in this
 section also include the verification of such a data integrity proof.
         </p>
 
+        <p class="advisement">
+When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is utilized,
+implementations MUST detect <a data-cite="RDF-CANON#dataset-poisoning">
+dataset poisoning</a> by default and abort processing.
+        </p>
+
         <section>
           <h4>Add Proof (eddsa-2022)</h4>
 
--- a/index.html
+++ b/index.html
@@ -403,7 +403,7 @@ section also include the verification of such a data integrity proof.
 
         <p class="advisement">
 When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is used,
-implementations MUST detect <a data-cite="RDF-CANON#dataset-poisoning">
+implementations will detect <a data-cite="RDF-CANON#dataset-poisoning">
 dataset poisoning</a> by default, and abort processing upon such detection.
         </p>
 


iherman (dismissed):
The text is normative, and I am not sure that it should be. Indeed, it refers to the RDFC 1.0 spec, but the normative definition of RDFC 1.0 already makes it a requirement to defend itself against poison graphs. See [§4.4.3](https://www.w3.org/TR/rdf-canon/#canon-algo-algo) which says, in the introduction:

> In some extreme cases, the algorihm can behave poorly, particularly in [Step 5](https://www.w3.org/TR/rdf-canon/#ca.5). Implementations MUST prevent against potenial denial-of-service attacks. See [7.1 Dataset Poisoning](https://www.w3.org/TR/rdf-canon/#dataset-poisoning) for futher information.

This means that the current steps introduces a normative requirement against an upcoming recommendation that has this requirement by itself.

Adding a note referring to the issue, possibly referring to the relevant section(s) of the RDFC 1.0 spec, and noting that the RDFC 1.0 might reject a specific graph seems to be enough.

msporny:
@iherman wrote:
> Adding a note referring to the issue, possibly referring to the relevant section(s) of the RDFC 1.0 spec, and noting that the RDFC 1.0 might reject a specific graph seems to be enough.

Yes, you make a good point. I'm happy to make that change across the specifications if @OR13 is ok with it. I believe he wanted us to speak normatively about this across all of the specifications, though agree that there is no need to because it is a normative requirement in RDFC1.

iherman:
> @iherman wrote:
> 
> > Adding a note referring to the issue, possibly referring to the relevant section(s) of the RDFC 1.0 spec, and noting that the RDFC 1.0 might reject a specific graph seems to be enough.
> 
> Yes, you make a good point. I'm happy to make that change across the specifications if @OR13 is ok with it. I believe he wanted us to speak normatively about this across all of the specifications, though agree that there is no need to because it is a normative requirement in RDFC1.

To be fair: this extra feature of RDFC 1.0 is very recent, the [relevant PR](https://github.com/w3c/rdf-canon/pull/138) in spec text was merged about a week ago, and the [PR on implementations and tests](https://github.com/w3c/rdf-canon/pull/131) three weeks ago. So I would not expect outsiders of the RCH to know about it…

Wind4Greg (approved):

dlongley (commented):
I agree with Ivan that this isn't needed. I'm not blocking either way, but I think the spec is simpler without it since it's already covered by the dependency.

OR13 (approved):
This seems like good advice to give implementers.

Implementations that don't abide the MUST are both non conformant and vulnerable to a trivial denial of service attack?

There may still be a denial of service attack that is not mitigated by implementing the underlying dependency correctly. I've not reviewed the specific algorithms, as they are currently defined or the proposed mitigations.

iherman:
> This seems like good advice to give implementers.
> 
> Implementations that don't abide the MUST are both non conformant and vulnerable to a trivial denial of service attack?
> 

There is a difference between a good advice and a normative _MUST_. The advice could just say that implementers should use conforming RDFC 1.0 implementations that will secure that DoS will not happen through canonicalization.

> There may still be a denial of service attack that is not mitigated by implementing the underlying dependency correctly. I've not reviewed the specific algorithms, as they are currently defined or the proposed mitigations.

Are you saying there may be DoS attacks that are independent of the RDFC 1.0 problem? Maybe, but I cannot really imagine one. This PR was explicitly referring to the poison datasets that can create problems with C14n. It should be part of a separate issue/PR if such is found.

dmitrizagidulin (changes requested):
I agree with @OR13 that this is good advice to give to implementers.
But I also agree with @iherman that the normative language is not quite appropriate here, especially because we now have _two_ upstream specs (RDFC and Data Integrity both) that make normative requirements of this.

I'd be +1 on this if we left the advice, just marked it non-normative. We can just point out, that RDFC and DI specs require this.

index.html
```diff
@@ -401,6 +401,12 @@ <h3>eddsa-2022</h3>
 section also include the verification of such a data integrity proof.
         </p>
 
+        <p class="advisement">
+When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is utilized,
+implementations MUST detect <a data-cite="RDF-CANON#dataset-poisoning">
```
msporny:
```suggestion
implementations will detect <a data-cite="RDF-CANON#dataset-poisoning">
```

Does this work for you @dlongley, @iherman, @dmitrizagidulin, and @OR13?

OR13: This works for me, if merged, i will approve.

index.html
```diff
@@ -401,6 +401,12 @@ <h3>eddsa-2022</h3>
 section also include the verification of such a data integrity proof.
         </p>
 
+        <p class="advisement">
+When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is utilized,
```
```suggestion
When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is used,
```

index.html
```diff
@@ -401,6 +401,12 @@ <h3>eddsa-2022</h3>
...
+implementations MUST detect <a data-cite="RDF-CANON#dataset-poisoning">
+dataset poisoning</a> by default and abort processing.
```
```suggestion
dataset poisoning</a> by default, and abort processing upon such detection.
```

TallTed (changes requested):
Small (I think, editorial) tweaks.

dmitrizagidulin (approved):

index.html
```diff
@@ -401,6 +401,12 @@ <h3>eddsa-2022</h3>
 section also include the verification of such a data integrity proof.
         </p>
 
+        <p class="advisement">
+When the RDF Dataset Canonicalization Algorithm [[RDF-CANON]] is used,
+implementations MUST detect <a data-cite="RDF-CANON#dataset-poisoning">
```
msporny:
```suggestion
implementations will detect <a data-cite="RDF-CANON#dataset-poisoning">
```

msporny:
I have updated this PR to remove the normative statement and align with the advisement approach as suggested by @iherman and @dmitrizagidulin (and agreed to by @OR13) in PR https://github.com/w3c/vc-di-ecdsa/pull/24.

msporny:
Editorial, multiple reviews, changes requested and made, no objections, merging.

Issue #66, opened 2023-11-03
Fix byte length of `publickKeyMultibase` to 34


JSAssassin:
See - https://github.com/w3c/vc-di-eddsa-test-suite/pull/21#discussion_r1380688724 & https://github.com/w3c/vc-di-eddsa-test-suite/pull/21#discussion_r1380838204

msporny:
PR #67 was raised to address this issue and merged. This issue is now resolved. Closing.

PR #67, opened 2023-11-03
Fix byte length of `publickKeyMultibase`.

JSAssassin:
Addresses #66

***
<a href="https://pr-preview.s3.amazonaws.com/JSAssassin/vc-di-eddsa/pull/67.html" title="Last updated on Nov 9, 2023, 9:33 PM UTC (0d9438f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/67/e95d3d7...JSAssassin:0d9438f.html" title="Last updated on Nov 9, 2023, 9:33 PM UTC (0d9438f)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -270,7 +270,7 @@ suites defined in this specification.
 
           <p>
 The `publicKeyMultibase` value of the verification method MUST be 34 bytes in
-length and starts with the base-58-btc prefix (`z`), as defined in the
+length and start with the base-58-btc prefix (`z`), as defined in the
 <a data-cite="VC-DATA-INTEGRITY#multibase-0">Multibase section</a> of
 [[VC-DATA-INTEGRITY]]. A Multibase-encoded Multikey value follows, which MUST
 consist of a binary value that starts with the two-byte prefix `0xed01`, which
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -269,7 +269,7 @@ <h4>Multikey</h4>
           </p>
 
           <p>
-The `publicKeyMultibase` value of the verification method MUST be 35 bytes in
+The `publicKeyMultibase` value of the verification method MUST be 34 bytes in
 length and starts with the base-58-btc prefix (`z`), as defined in the
```
```suggestion
length and start with the base-58-btc prefix (`z`), as defined in the
```

index.html
```diff
@@ -1384,7 +1384,7 @@ <h4>Ed25519VerificationKey2020</h4>
             </p>
 
             <p>
-  The `publicKeyMultibase` value of the verification method MUST be 35 bytes in
+  The `publicKeyMultibase` value of the verification method MUST be 34 bytes in
   length and starts with the base-58-btc prefix (`z`), as defined in the
```
```suggestion
  length and start with the base-58-btc prefix (`z`), as defined in the
```

Wind4Greg (commented):
Hmm, it seems that the description is also bit problematic. The `publicKeyMultibase` value is a **string** that contains the base-58-btc encoding of the 2 byte prefix plus the 32 byte key. So the length in bytes of what is being encoded is 34 bytes but the string size is whatever. Shouldn't we re-write this a little to be clearer.

msporny (approved):

msporny:
> Hmm, it seems that the description is also bit problematic. The `publicKeyMultibase` value is a **string** that contains the base-58-btc encoding of the 2 byte prefix plus the 32 byte key. So the length in bytes of what is being encoded is 34 bytes but the string size is whatever. Shouldn't we re-write this a little to be clearer.

Yes, I'll do some editorial fix-ups to make this more clear after the merge. I can't edit the PR, so I have to do the fixes in a separate commit.

UPDATE: This is now complete in commit 215cdbad91a30ac40f4456bd625451569e26ed61

msporny:
Arguably normative, multiple reviews, changes requested (and will be made in a separate commit), no objections, merging.

PR #79, opened 2024-05-01
JCS Context Injection

Wind4Greg:
This PR addresses issue https://github.com/w3c/vc-data-integrity/issues/225 to properly handle `@context` injection issues that could arise with JCS. Note that this also applies to `@context` issues that could arise with proof sets/chains with JCS. This PR contains **normative** changes to the *Create Proof (eddsa-jcs-2022)* and *Verify Proof (eddsa-jcs-2022)* algorithms.

This PR does **NOT** include **informative** text such as:

1. Explanation of why these steps are necessary and sufficient. Such informative text is present in the discussions of issue  https://github.com/w3c/vc-data-integrity/issues/225 and could be extracted and included somewhere in the document. Thoughts?
2. Updates to the current *eddsa-jcs-2022* test vectors in the specification. Will be updated after this PR is reviewed.
3. Explicit test vectors illustrating these procedures in a realistic example. Is this desirable/necessary? Thoughts

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-di-eddsa/pull/79.html" title="Last updated on Jun 9, 2024, 4:07 PM UTC (c49c133)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/79/0944119...Wind4Greg:c49c133.html" title="Last updated on Jun 9, 2024, 4:07 PM UTC (c49c133)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -886,13 +886,13 @@ JSON Canonicalization Scheme [[RFC8785]] to the <var>proofConfig</var>.
 Let |transformedData| be the result of running the algorithm in Section
 <a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument|,
 |proofConfig|, and |options| passed as parameters where step
-<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
-text respectively:
+<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
+the following texts:
                 <ol class="algorithm">
                   <li>
 If <var>options</var>.<var>type</var> is not set to the string
 `DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
-set to the string `eddsa-jcs-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
+set to the string `eddsa-jcs-2022`, then a `PROOF_TRANSFORMATION_ERROR` MUST be
 raised.
                   </li>
                   <li>
@@ -965,13 +965,13 @@ the |proofOptions|.<var>@context</var>:
               <ol class="algorithm">
                 <li>
 The |securedDocument|.<var>@context</var> MUST contain all values contained in the
-|proofOptions|.<var>@context</var> in the same order. Otherwise set |verified| to
+|proofOptions|.<var>@context</var> in the same order. Otherwise, set |verified| to
 `false` and skip to the last step.
                 </li>
                 <li>
 The |securedDocument|.<var>@context</var> MUST NOT have any values preceding the
-values contained in the |proofOptions|.<var>@context</var> but those MAY be followed by
-additional values not contained in the |proof|.<var>@context</var>. Otherwise
+values contained in the |proofOptions|.<var>@context</var>, but those MAY be followed by
+additional values not contained in the |proof|.<var>@context</var>. Otherwise,
 set |verified| to `false` and skip to the last step.
                 </li>
               </ol>
@@ -985,13 +985,13 @@ Set |securedDocument|.<var>@context</var> equal to
 Let |transformedData| be the result of running the algorithm in Section
 <a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument| and
 |proofOptions| passed as parameters where step
-<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
-text respectively:
+<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
+the following texts:
                 <ol class="algorithm">
                   <li>
 If <var>options</var>.<var>type</var> is not set to the string
 `DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
-set to the string `eddsa-jcs-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
+set to the string `eddsa-jcs-2022`, then a `PROOF_TRANSFORMATION_ERROR` MUST be
 raised.
                   </li>
                   <li>
@@ -1005,7 +1005,7 @@ Let |proofConfig| be the result of running the algorithm in
 Section [[[#proof-configuration-eddsa-rdfc-2022]]] with
 |unsecuredDocument| and |proofOptions| passed as a parameters and where
 steps <strong>4)</strong> and
-<strong>9)</strong> are replaced by the following text respectively:
+<strong>9)</strong> are respectively replaced by the following texts:
                 <ol class="algorithm">
                   <li>
 If <var>options</var>.<var>type</var> is not set to
@@ -1035,7 +1035,7 @@ JSON Canonicalization Scheme [[RFC8785]] to the <var>proofConfig</var>.
                   <dd>|verified|</dd>
                   <dt>[=verifiedDocument=]</dt>
                   <dd>
-  |unsecuredDocument| if |verified| is `true`, otherwise <a data-cite="INFRA#nulls">Null</a></dd>
+  |unsecuredDocument|, if |verified| is `true`; otherwise, <a data-cite="INFRA#nulls">Null</a></dd>
                 </dl>
               </li>
             </ol>
--- a/index.html
+++ b/index.html
@@ -866,50 +866,22 @@ Let |proof| be a clone of the proof options, |options|.
               </li>
               <li>
 Let |proofConfig| be the result of running the algorithm in
-Section [[[#proof-configuration-eddsa-rdfc-2022]]] with
-|options| passed as a parameter and where
-steps <strong>4)</strong> and
-<strong>9)</strong> are replaced by the following text respectively:
-                <ol class="algorithm">
-                  <li>
-If <var>options</var>.<var>type</var> is not set to
-`DataIntegrityProof` and <var>proofConfig</var>.<var>cryptosuite</var> is not
-set to `eddsa-jcs-2022`, an error MUST be raised and it SHOULD use error
-code `PROOF_GENERATION_ERROR`.
-                  </li>
-                  <li>
-Let <var>canonicalProofConfig</var> be the result of applying the
-JSON Canonicalization Scheme [[RFC8785]] to the <var>proofConfig</var>.
-                  </li>
-                </ol>
+Section [[[#proof-configuration-eddsa-jcs-2022]]] with
+|options| passed as a parameter.
               </li>
               <li>
 Let |transformedData| be the result of running the algorithm in Section
-<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument|,
-|proofConfig|, and |options| passed as parameters where step
-<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
-the following texts:
-                <ol class="algorithm">
-                  <li>
-If <var>options</var>.<var>type</var> is not set to the string
-`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
-set to the string `eddsa-jcs-2022`, then an error MUST be raised and it SHOULD
-use error code `PROOF_GENERATION_ERROR`.
-                  </li>
-                  <li>
-Let <var>canonicalDocument</var> be the result of applying the
-JSON Canonicalization Scheme [[RFC8785]] to the <var>unsecuredDocument</var>.
-                  </li>
-                </ol>
+<a href="#transformation-eddsa-jcs-2022"></a> with |unsecuredDocument|,
+|proofConfig|, and |options| passed as parameters.
               </li>
               <li>
 Let |hashData| be the result of running the algorithm in Section
-[[[#hashing-eddsa-rdfc-2022]]] with |transformedData| and |proofConfig|
+[[[#hashing-eddsa-jcs-2022]]] with |transformedData| and |proofConfig|
 passed as a parameters.
               </li>
               <li>
 Let |proofBytes| be the result of running the algorithm in Section
-[[[#proof-serialization-eddsa-rdfc-2022]]] with |hashData| and
+[[[#proof-serialization-eddsa-jcs-2022]]] with |hashData| and
 |options| passed as parameters.
               </li>
               <li>
@@ -976,49 +948,22 @@ Set |unsecuredDocument|.<var>@context</var> equal to
             </li>
             <li>
 Let |transformedData| be the result of running the algorithm in Section
-<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument| and
-|proofOptions| passed as parameters where step
-<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
-the following texts:
-                <ol class="algorithm">
-                  <li>
-If <var>options</var>.<var>type</var> is not set to the string
-`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
-set to the string `eddsa-jcs-2022`, then a `PROOF_TRANSFORMATION_ERROR` MUST be
-raised.
-                  </li>
-                  <li>
-Let <var>canonicalDocument</var> be the result of applying the
-JSON Canonicalization Scheme [[RFC8785]] to the <var>unsecuredDocument</var>.
-                  </li>
-                </ol>
+<a href="#transformation-eddsa-jcs-2022"></a> with |unsecuredDocument| and
+|proofOptions| passed as parameters.
               </li>
             <li>
 Let |proofConfig| be the result of running the algorithm in
-Section [[[#proof-configuration-eddsa-rdfc-2022]]] with
-|unsecuredDocument| and |proofOptions| passed as a parameters and where step
-<strong>8)</strong> is not performed and steps <strong>4)</strong> and
-<strong>9)</strong> are respectively replaced by the following texts:
-                <ol class="algorithm">
-                  <li>
-If <var>options</var>.<var>type</var> is not set to
-`DataIntegrityProof` and <var>proofConfig</var>.<var>cryptosuite</var> is not
-set to `eddsa-jcs-2022`, an `INVALID_PROOF_CONFIGURATION` error MUST be raised.
-                  </li>
-                  <li>
-Let <var>canonicalProofConfig</var> be the result of applying the
-JSON Canonicalization Scheme [[RFC8785]] to the <var>proofConfig</var>.
-                  </li>
-                </ol>
-              </li>
+Section [[[#proof-configuration-eddsa-jcs-2022]]] with
+|unsecuredDocument| and |proofOptions| passed as a parameters.
+            </li>
             <li>
   Let |hashData| be the result of running the algorithm in Section
-  [[[#hashing-eddsa-rdfc-2022]]] with |transformedData| and |proofConfig|
+  [[[#hashing-eddsa-jcs-2022]]] with |transformedData| and |proofConfig|
   passed as a parameters.
             </li>
             <li>
   Let |verified:boolean| be the result of running the algorithm in Section
-  [[[#proof-verification-eddsa-rdfc-2022]]] algorithm on |hashData|,
+  [[[#proof-verification-eddsa-jcs-2022]]] algorithm on |hashData|,
   |proofBytes|, and |proofConfig|.
               </li>
               <li>
@@ -1257,7 +1202,7 @@ Return <var>verificationResult</var> as the <em>verification result</em>.
 
         </section>
       </section>
-<!-- End of JCS section -->
+
     </section>
 
     <section>
--- a/index.html
+++ b/index.html
@@ -841,45 +841,10 @@ Return <var>verificationResult</var> as the <em>verification result</em>.
 The `eddsa-jcs-2022` cryptographic suite takes an input document, canonicalizes
 the document using the JSON Canonicalization Scheme [[RFC8785]], and then
 cryptographically hashes and signs the output resulting in the production of a
-data integrity proof. The algorithms for this cryptographic suite are the same
-as the ones in Section <a href="#eddsa-rdfc-2022"></a> except for the
-following modifications:
+data integrity proof.
           </p>
 
-          <p>
-In Section <a href="#transformation-eddsa-rdfc-2022"></a>, step
-<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
-text:
-          </p>
 
-          <ol class="algorithm">
-            <li>
-If <var>options</var>.<var>type</var> is not set to the string
-`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
-set to the string `eddsa-jcs-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
-raised.
-            </li>
-            <li>
-Let <var>canonicalDocument</var> be the result of applying the
-JSON Canonicalization Scheme [[RFC8785]] to the <var>unsecuredDocument</var>.
-            </li>
-          </ol>
-
-          <p>
-In Section <a href="#proof-configuration-eddsa-rdfc-2022"></a>, step
-<strong>8)</strong> is not performed, and steps <strong>4)</strong> and
-<strong>9)</strong> are replaced by the following text:
-          </p>
-
-          <p style="padding-left: 2em;">
-<strong>4)</strong> If <var>options</var>.<var>type</var> is not set to
-`DataIntegrityProof` and <var>proofConfig</var>.<var>cryptosuite</var> is not
-set to `eddsa-jcs-2022`, an `INVALID_PROOF_CONFIGURATION` error MUST be raised.
-          </p>
-          <p style="padding-left: 2em;">
-<strong>9)</strong> Let <var>canonicalProofConfig</var> be the result of applying the
-JSON Canonicalization Scheme [[RFC8785]] to the <var>proofConfig</var>.
-          </p>
 
           <section>
             <h4>Create Proof (eddsa-jcs-2022)</h4>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -961,21 +961,25 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
             </li>
             <li>
 Perform the following checks  between the |securedDocument|.<var>@context</var> and
-the |proof|.<var>@context</var>:
+the |proofOptions|.<var>@context</var>:
               <ol class="algorithm">
                 <li>
 The |securedDocument|.<var>@context</var> MUST contain all values contained in the
-|proof|.<var>@context</var> in the same order. Otherwise set |verified| to
+|proofOptions|.<var>@context</var> in the same order. Otherwise set |verified| to
```
```suggestion
|proofOptions|.<var>@context</var> in the same order. Otherwise, set |verified| to
```

index.html
```diff
@@ -961,21 +961,25 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
 `false` and skip to the last step.
                 </li>
                 <li>
 The |securedDocument|.<var>@context</var> MUST NOT have any values preceding the
-values contained in the |proof|.<var>@context</var> but those MAY be followed by
+values contained in the |proofOptions|.<var>@context</var> but those MAY be followed by
 additional values not contained in the |proof|.<var>@context</var>. Otherwise
```
```suggestion
additional values not contained in the |proof|.<var>@context</var>. Otherwise,
```

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+  |proofBytes|, and |proofConfig|.
+              </li>
+              <li>
+  Return a [=verification result=] with [=struct/items=]:
+                <dl data-link-for="verification result">
+                  <dt>[=verified=]</dt>
+                  <dd>|verified|</dd>
+                  <dt>[=verifiedDocument=]</dt>
+                  <dd>
+  |unsecuredDocument| if |verified| is `true`, otherwise <a data-cite="INFRA#nulls">Null</a></dd>
```
```suggestion
  |unsecuredDocument|, if |verified| is `true`; otherwise, <a data-cite="INFRA#nulls">Null</a></dd>
```

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+JSON Canonicalization Scheme [[RFC8785]] to the <var>unsecuredDocument</var>.
+                  </li>
+                </ol>
+              </li>
+            <li>
+Let |proofConfig| be the result of running the algorithm in
+Section [[[#proof-configuration-eddsa-rdfc-2022]]] with
+|unsecuredDocument| and |proofOptions| passed as a parameters and where
+steps <strong>4)</strong> and
+<strong>9)</strong> are replaced by the following text respectively:
```
```suggestion
<strong>9)</strong> are respectively replaced by the following texts:
```

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+Let |transformedData| be the result of running the algorithm in Section
+<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument| and
+|proofOptions| passed as parameters where step
+<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
+text respectively:
+                <ol class="algorithm">
+                  <li>
+If <var>options</var>.<var>type</var> is not set to the string
+`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
+set to the string `eddsa-jcs-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
```
```suggestion
set to the string `eddsa-jcs-2022`, then a `PROOF_TRANSFORMATION_ERROR` MUST be
```

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+            <li>
+Set |securedDocument|.<var>@context</var> equal to
+|proofOptions|.<var>@context</var>.
+            </li>
+            <li>
+Let |transformedData| be the result of running the algorithm in Section
+<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument| and
+|proofOptions| passed as parameters where step
+<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
+text respectively:
```
```suggestion
<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
the following texts:
```

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+the |proofOptions|.<var>@context</var>:
+              <ol class="algorithm">
+                <li>
+The |securedDocument|.<var>@context</var> MUST contain all values contained in the
+|proofOptions|.<var>@context</var> in the same order. Otherwise set |verified| to
+`false` and skip to the last step.
+                </li>
+                <li>
+The |securedDocument|.<var>@context</var> MUST NOT have any values preceding the
+values contained in the |proofOptions|.<var>@context</var> but those MAY be followed by
```
```suggestion
values contained in the |proofOptions|.<var>@context</var>, but those MAY be followed by
```

index.html
```diff
@@ -892,11 +860,69 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
+Let |transformedData| be the result of running the algorithm in Section
+<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument|,
+|proofConfig|, and |options| passed as parameters where step
+<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
+text respectively:
+                <ol class="algorithm">
+                  <li>
+If <var>options</var>.<var>type</var> is not set to the string
+`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
+set to the string `eddsa-jcs-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
```
```suggestion
set to the string `eddsa-jcs-2022`, then a `PROOF_TRANSFORMATION_ERROR` MUST be
```

index.html
```diff
@@ -892,11 +860,69 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
+JSON Canonicalization Scheme [[RFC8785]] to the <var>proofConfig</var>.
+                  </li>
+                </ol>
+              </li>
+              <li>
+Let |transformedData| be the result of running the algorithm in Section
+<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument|,
+|proofConfig|, and |options| passed as parameters where step
+<strong>1)</strong> and step <strong>2)</strong> are replaced by the following
+text respectively:
```
```suggestion
<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
the following texts:
```

TallTed (changes requested):
punctuation, word order, and the like

rflechtner:
Very cool, thank you! Makes perfect sense as far as I'm concerned, and I appreciate a second pair of eyes on the proposed solution. I also don't think we'd need additional test vectors here, however I do think we should cover all edge cases in https://github.com/w3c/vc-di-eddsa-test-suite to make sure the logic is sound and does not introduce other unforeseen issues. Happy to contribute here, but I may need some onboarding as I'm struggling to understand in how far the test suite represents the current state of all relevant specifications (e.g., w/r/ to did:key & Multikey encodings, https://github.com/w3c/vc-di-eddsa-test-suite/issues/44).

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+value</a> in |securedDocument|.|proof|.|proofValue|.
+            </li>
+            <li>
+Perform the following checks  between the |securedDocument|.<var>@context</var> and
+the |proofOptions|.<var>@context</var>:
+              <ol class="algorithm">
+                <li>
+The |securedDocument|.<var>@context</var> MUST contain all values contained in the
+|proofOptions|.<var>@context</var> in the same order. Otherwise, set |verified| to
+`false` and skip to the last step.
```
silverpill: This change would make all existing secured documents invalid. Therefore, the name of the cryptosuite should be changed as well (to `eddsa-jcs-2024`, for example).

rflechtner: It's designed not to invalidate existing proofs. Existing proofs should currently not contain a @context property, which means that this rule will have no effect on them. But maybe we can clarify in the rule that if there is no @context on the proof (which is equivalent to an empty set of contexts), no checks on the document context are performed.

rflechtner: But if you have concerns regarding the logic of this or know of existing implementations that for some reason do add @context values to the proofs emitted please let us know, I can't guarantee that the solution considers everything there is to consider

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+values contained in the |proofOptions|.<var>@context</var>, but those MAY be followed by
+additional values not contained in the |proof|.<var>@context</var>. Otherwise,
+set |verified| to `false` and skip to the last step.
+                </li>
+              </ol>
+
+            </li>
+            <li>
+Set |securedDocument|.<var>@context</var> equal to
+|proofOptions|.<var>@context</var>.
```
silverpill: What action should be taken if `proofOptions` doesn't have `@context`?

index.html
silverpill: >Existing proofs should currently not contain a `@context` property, which means that this rule will have no effect on them. 

Oh, good to know.

>But maybe we can clarify in the rule that if there is no `@context` on the proof (which is equivalent to an empty set of contexts), no checks on the document context are performed.

Yes, I think this can be clarified. Currently it sounds like `@context` is required: https://github.com/w3c/vc-di-eddsa/pull/79#discussion_r1587597825

I don't know of any implementations that add `@context` to `proofOptions`

index.html
rflechtner: Yes, that is an important point: in this case, we need to assume that we're in 'compatibility mode' for support of proofs generated according to the specification as it existed prior to this change, and _NOT_ touch the document's @context at all

Wind4Greg: Hi all (@rflechtner, @silverpill, @dlongley, @msporny ), in making this update, for simplicity and security, I put back step 8 from [Proof Configuration (eddsa-rdfc-2022)](https://w3c.github.io/vc-di-eddsa/#proof-configuration-eddsa-rdfc-2022) which copies the existing document `@context` into *proofConfig* prior to canonicalization and then hashing.  This further protects the contents of the *proof* field, but also needs to be accounted for to realize backward compatibility (results in different hash values being signed).

It seems that all algorithm tweaks can be detected and compensated for via the presence/absence of the *proof.@context* field. 

**Discussion**: Do we want to add these complications? How do we "officially" talk (or not) about this backwards compatibility? Would greatly appreciate help on informative text for (a) this approach to handling context with JCS, and (b) backwards compatibility.

Note my background is more from the cryptography side rather than JSON-LD side and I've been generating most of the test vectors for the specs so I'm used to implementing/tweaking the algorithms.  Also once we get this figured out for EdDSA I was planning on updating the ECDSA JCS stuff too. Cheers Greg

silverpill: Defining a new cryptosuite might still be worth it, if more normative changes are planned (https://github.com/w3c/vc-data-integrity/issues/231; potential problems with nested secured documents?)

rflechtner: > in making this update, for simplicity and security, I put back step 8 from [Proof Configuration (eddsa-rdfc-2022)](https://w3c.github.io/vc-di-eddsa/#proof-configuration-eddsa-rdfc-2022) which copies the existing document `@context` into _proofConfig_ prior to canonicalization and then hashing. This further protects the contents of the _proof_ field, but also needs to be accounted for to realize backward compatibility (results in different hash values being signed).

Thanks for pointing this out, I had missed this detail. Indeed, writing the document context to the proof context during verification would result in a loss of interoperability with existing implementations, which I think was a goal. I'd argue we would thus either have to explicitly disallow step 8 _only on verification_, or bump the cryptosuite version as @silverpill suggested (https://github.com/w3c/vc-di-eddsa/pull/79#discussion_r1587384419), in which case we do not need to consider backwards compatibility. Don't know if the latter is on the table though, as I'm not too involved with the WG. @dlongley & @msporny would be better suited to make a call here.

As for discussing reasons for these steps, I think that would help avoiding implementation errors, as instructions like 'omit step 8' that are not immediately plausible are more easily overlooked.

Wind4Greg: Hi @silverpill and @rflechtner, coming up with a new cryptosuite is a significant burden. To remain "compatible" with the "previous versions" of the procedures involving JCS (EdDSA and ECDSA) would just require a simple check for the *proof.@context* field and some slightly different processing if its not present. This seems the best way to go to me.

That is we would be able to verify previously signed JCS based VC, but we would specify that all (new) VCs based on JCS would only use the new procedures (since those VCs are tolerant to injection, can be used in proof chains/sets, etc...).

Thoughts? If you concur I can enhance the verification procedure appropriately and add a test vector showing this.

index.html
```diff
@@ -892,11 +860,69 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
+Let |proofConfig| be the result of running the algorithm in
+Section [[[#proof-configuration-eddsa-rdfc-2022]]] with
+|options| passed as a parameter and where
+steps <strong>4)</strong> and
+<strong>9)</strong> are replaced by the following text respectively:
+                <ol class="algorithm">
+                  <li>
+If <var>options</var>.<var>type</var> is not set to
+`DataIntegrityProof` and <var>proofConfig</var>.<var>cryptosuite</var> is not
+set to `eddsa-jcs-2022`, an `INVALID_PROOF_CONFIGURATION` error MUST be raised.
```
dlongley: Two points: `INVALID_PROOF_CONFIGURATION` has been removed from the data integrity spec so we need a new error code here to being that inline and we the error type/code to be a SHOULD. We can do this in a separate PR if necessary -- we'll need the same fix to the other suites here.

```suggestion
`DataIntegrityProof` and <var>proofConfig</var>.<var>cryptosuite</var> is not
set to `eddsa-jcs-2022`, an error MUST be raised and it SHOULD use error
code `PROOF_GENERATION_ERROR`.
```

index.html
```diff
@@ -892,11 +860,69 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
+<a href="#transformation-eddsa-rdfc-2022"></a> with |unsecuredDocument|,
+|proofConfig|, and |options| passed as parameters where step
+<strong>1)</strong> and step <strong>2)</strong> are respectively replaced by
+the following texts:
+                <ol class="algorithm">
+                  <li>
+If <var>options</var>.<var>type</var> is not set to the string
+`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
+set to the string `eddsa-jcs-2022`, then a `PROOF_TRANSFORMATION_ERROR` MUST be
+raised.
```
dlongley:
```suggestion
set to the string `eddsa-jcs-2022`, then an error MUST be raised and it SHOULD
use error code `PROOF_GENERATION_ERROR`.
```

index.html
```diff
@@ -919,11 +945,100 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+removed.
+            </li>
+            <li>
+Let |proofBytes| be the
+<a data-cite="VC-DATA-INTEGRITY#multibase-0">Multibase decoded base58-btc
+value</a> in |securedDocument|.|proof|.|proofValue|.
+            </li>
+            <li>
+Perform the following checks  between the |securedDocument|.<var>@context</var> and
+the |proofOptions|.<var>@context</var>:
```
dlongley: This should give us backwards compatibility.

Also note that if a holder deletes `proof.@context`, and `securedDocument.@context` has changed since the time of signing, then the signature will fail since the signed canonicalized proof will differ when `securedDocument.@context` is still copied over to `proof.@context` to generate the canonical proof to sign.

```suggestion
If |proofOptions|.<var>@context</var> exists, perform the following checks  between the |securedDocument|.<var>@context</var> the |proofOptions|.<var>@context</var>:
```

index.html
dlongley: Can we simplify to this and remove the second MUST NOT?:

```suggestion
The |securedDocument|.<var>@context</var> MUST start with all values contained in the
|proofOptions|.<var>@context</var> in the same order. Otherwise, set |verified| to
`false` and skip to the last step.
```

index.html
dlongley: I think we can move this "Set..." step into the "If" conditional I suggested above for backwards compatibility.

dlongley (changes requested):

index.html
dlongley: The same protection should be afforded even if `@context` isn't copied to `proof` (for proof canonicalization) in this backwards compatibility case because the document signature would fail if its `@context` had changed. So the carve out @Wind4Greg mentions here: https://github.com/w3c/vc-di-eddsa/pull/79#discussion_r1589688204 should also work (and it seems work better).

index.html
dlongley: It seems that if we delete adding the context within the JCS proof configuration algorithm, then we can make these two additional changes:

1. During proof generation, the unsecured document context should be added to the (cloned) proof options before running the proof configuration algorithm. This will cause the same effect as having the proof configuration algorithm add it.
2. During proof verification, if `@context` is present in the proof options, ensure the secured document `@context` values start with its value and then set the unsecured document `@context` to the value from the proof options (no modification should be done to the secured document). This will cause no `@context` to be present later in the proof configuration algorithm for the backwards compatibility case (because nothing will add it). If the document context has changed, verification will fail as expected -- as the backwards compatible case doesn't support proof sets/chains. For the regular case, `@context` will be present because it was in the proof options (coming from the `proof` in the document during verification). If a holder removes `@context` from the proof in the regular case, then it will fail to verify because it was present when signed.

Wind4Greg: Hi @dlongley, we don't have a separate "JCS Proof Configuration" algorithm. So I implemented your above suggestion just by modifying the use of the "RDF Proof Configuration" algorithm in the verification step. In addition, I cleaned up some mistakes where *securedDocument* should have been *unsecuredDocument*. Please review again.

Wind4Greg: Hi all, just updated test vector code with the compatibility checks discussed here. See https://github.com/Wind4Greg/EdDSA-Test-Vectors/blob/jcs-injection/JCSDataIntegrityVerify.js and was able to verify both old and new JCS EdDSA test vectors. Cheers Greg

silverpill (commented):
LGTM

dlongley (approved):

Wind4Greg:
@silverpill can you formally click the "approve" button if you are okay with this PR. Then we can work to get it merged. @rflechtner how does it look to you? Once we get this done. I'll mirror these changes to the ECDSA document. Cheers Greg

silverpill (approved):

rflechtner (approved):
Sorry for the delay, I wanted to try a tentative implementation before giving feedback. The solution seems functional and reasonable to me though, very happy with this change.
There's two places where implementers may have to read twice in order to get it right:
- The document context is effectively added to the proof / proofOptions twice during proof creation, once as part of proof configuration and once in step 6.
- For proof verification, "where step 8) is not performed" is easily missed - maybe an emphasis helps here?

Wind4Greg:
Hi @rflechtner, whether to write things for JCS as a "delta" on the RDFC or as new procedures has been a question and an inconsistency across the specs. I just put in a PR for JCS injection for the ECDSA case. Take a look at https://github.com/w3c/vc-di-ecdsa/pull/61. In that document (VC-DI-ECDSA) there were separate sub-procedures for JCS. Preferences?

rflechtner:
> Preferences?

Yeah I do have a clear preference for having separate sub-procedures. While there's certainly duplication it's just so much easier to understand and implement.
Writing it as a diff is cognitively quite demanding because you need to jump between two different procedures and then try to patch the resulting routine together in your head from these different instructions.

Maybe the example is a little far-fetched, but in German legislation proposals for changes to existing laws are now more and more required to include a 'synopsis' that shows you the before and after, because even professionals had problems following changes written as 'in paragraph x replace formulation y with x'.

Wind4Greg:
Hi all, (@silverpill, @rflechtner, @dlongley, @TallTed), I'm going to update this PR to use separate "sub-procedures" (rather than deltas) to match what was done in the [ECDSA Spec](https://w3c.github.io/vc-di-ecdsa/#ecdsa-jcs-2019) for JCS based proofs. I'm pretty sure that was a more recent write up and it would be nice for implementers if these looked the same (the only real different is the signature algorithm).

Wind4Greg:
Hi all, (@silverpill, @rflechtner, @dlongley, @TallTed), updated this PR to use separate "sub-procedures" (rather than deltas) to match what was done in the [ECDSA Spec](https://w3c.github.io/vc-di-ecdsa/#ecdsa-jcs-2019) for JCS based proofs. So, I'm asking for a re-review. Greg

rflechtner (commented):
We now have the issue again that proof configuration sets the proof `@context` on verification, which breaks the backwards compatibility case. Generally I think we should get rid of step 8 of the proof configuration algorithm, as @dlongley suggested (https://github.com/w3c/vc-di-eddsa/pull/79#discussion_r1589760761), and instead add the context to `options` as the very first step of the create proof algorithm. This makes it easier to understand and fixes the backwards compatibility case.

rflechtner:
Thanks for the re-write though, I think this is indeed an improvement!

Wind4Greg:
Looks like the last update lost a key change. Will fix.

Wind4Greg:
Hi @rflechtner updated. Please review carefully and also take a look at the ECDSA PR https://github.com/w3c/vc-di-ecdsa/pull/61. All these sub-procedures can get confusing. My test vector generation code is pretty flat so easier to check. Cheers Greg

silverpill (approved):

index.html
```diff
@@ -892,11 +860,42 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
-[[[#create-proof-eddsa-rdfc-2022]]] with the changes outlined in
-Section [[[#eddsa-jcs-2022]]].
-            </p>
+            <ol class="algorithm">
+              <li>
+Set <var>options</var>.<var>@context</var> to
+<var>unsecuredDocument</var>.<var>@context</var>.
+              </li>
+              <li>
+Let |proof| be a clone of the proof options, |options|.
```
dlongley: I'm somewhat inclined to say nothing about this, but will mention a minor quibble here, which is that someone might complain / ask whether they can make a copy of options with this modification instead of modifying it directly. They can of course do this. Maybe others do not see this as a problem -- and if so, please ignore.

If it is a problem, one way to address this is to add something like `optionsCopy` (or set `options` to a copy of itself) and instead modify that (to add `@context`) and then make sure that is what is passed in all the subsequent steps. We could also perhaps just flip these steps and then modify the language where `options` is passed below in various steps to pass `proof` in its place as "options", if that makes sense.

Again, maybe this is a non-issue. We could perhaps leave a note that making a copy of `options` before modifying it as long as you pass the copy along is fine.

index.html
```diff
@@ -892,11 +860,42 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
+              </li>
+              <li>
+Let |proofConfig| be the result of running the algorithm in
+Section [[[#proof-configuration-eddsa-jcs-2022]]] with
+|options| passed as a parameter.
+              </li>
+              <li>
+Let |transformedData| be the result of running the algorithm in Section
+<a href="#transformation-eddsa-jcs-2022"></a> with |unsecuredDocument|,
+|proofConfig|, and |options| passed as parameters.
```
dlongley: This just takes `options`, not `proofConfig`, right?

```suggestion
<a href="#transformation-eddsa-jcs-2022"></a> with |unsecuredDocument|
and |options| passed as parameters.
```

index.html
```diff
@@ -919,13 +918,285 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+produced as output. Whenever this algorithm encodes strings, it MUST use UTF-8
+encoding.
+          </p>
+
+          <ol class="algorithm">
+            <li>
+If <var>options</var>.<var>type</var> is not set to the string
+`DataIntegrityProof` and <var>options</var>.<var>cryptosuite</var> is not
+set to the string `eddsa-jcs-2022` then a `PROOF_TRANSFORMATION_ERROR` MUST be
+raised.
```
dlongley: Errors are MUSTs, specific error codes are SHOULDs, and `PROOF_TRANSFORMATION_ERROR` isn't defined anymore, so I picked `MALFORMED_PROOF_ERROR`:

```suggestion
set to the string `eddsa-jcs-2022` then an error MUST be raised that
SHOULD use the `MALFORMED_PROOF_ERROR` error code.
```

index.html
```diff
@@ -919,13 +918,285 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+for the
+<a data-cite="vc-data-integrity#dfn-cryptosuite">
+cryptographic suite</a> (<var>type</var>) and MUST contain a cryptosuite
+identifier (<var>cryptosuite</var>). A <em>proof configuration</em>
+object is produced as output.
+          </p>
+
+          <ol class="algorithm">
+            <li>
+Let <var>proofConfig</var> be an empty object.
```
dlongley: This looks a little dated or like it didn't receive the same fixes as ECDSA. The `proofConfig` could be a clone of options and then checks should be performed on specific properties.

```suggestion
Let <var>proofConfig</var> be a clone of the <var>options</var> object.
```

index.html
```diff
@@ -919,13 +918,285 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+            </li>
+            <li>
+Set <var>proofConfig</var>.<var>type</var> to
+<var>options</var>.<var>type</var>.
+            </li>
+            <li>
+If <var>options</var>.<var>cryptosuite</var> is set, set
+<var>proofConfig</var>.<var>cryptosuite</var> to its value.
+            </li>
```
dlongley: See above, I think we can cut these due to the cloning step.

```suggestion
```

index.html
```diff
@@ -919,13 +918,285 @@ <h4>Verify Proof (eddsa-jcs-2022)</h4>
...
+            <li>
+If <var>options</var>.<var>type</var> is not set to `DataIntegrityProof` and
+<var>proofConfig</var>.<var>cryptosuite</var> is not set to `eddsa-jcs-2022`, an
+`INVALID_PROOF_CONFIGURATION` error MUST be raised.
```
dlongley, line 1087: This needs to become `MALFORMED_PROOF_ERROR` and use similar language to the above text. We probably also just want to run the checks on `proofConfig` since it's now a clone. Similar to this:

https://w3c.github.io/vc-di-ecdsa/#proof-configuration-ecdsa-rdfc-2019

Though the text at that link still needs the error language updates / reconciliation with what was fixed with the interface in the data integrity core spec. If necessary, we can just do those pieces in a separate PR, but we do need to bring the error language inline.

dlongley (changes requested):
A few things we need to do here that weren't really part of the original task of doing a full algorithm replacement here vs. referencing w/in-place edits, but we do need to do them at some point (this is around reconciling the interface and error code changes in the core DI spec). Your call @Wind4Greg if you want to do them here or in a follow on PR.

index.html
Wind4Greg: Let me "flip the steps" a bit. Most JavaScript programmers have been burned by unintentional modifications to passed in objects and arrays to not like the "smell" of the current text. Good catch.

index.html
Wind4Greg: Yes. I'll fix this with the previous fix.

index.html
Wind4Greg: I'm a bit confused the link above has a `INVALID_PROOF_CONFIGURATION` error. Is that supposed to be a `MALFORMED_PROOF_ERROR`? Then there is language changes for error handling that needed to be check for all specifications right? Maybe get this PR done then scrub through everything, unless this is the only change needed.

Wind4Greg: I forgot is the error language in the DI spec the latest, e.g. "an error MUST be raised and SHOULD convey an error type of [PROOF_GENERATION_ERROR](https://w3c.github.io/vc-data-integrity/#PROOF_GENERATION_ERROR)."

And we want to make sure all cryptosuite documents conform and use listed error codes?

index.html
```diff
@@ -862,11 +862,11 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
-<var>unsecuredDocument</var>.<var>@context</var>.
+Let |proof| be a clone of the proof options, |options|.
               </li>
               <li>
-Let |proof| be a clone of the proof options, |options|.
+Set <var>proof</var>.<var>@context</var> to
+<var>unsecuredDocument</var>.<var>@context</var>.
               </li>
               <li>
 Let |proofConfig| be the result of running the algorithm in
```
dlongley, line 872: `proof` has to be passed instead of `options` here to properly canonicalize it with the added `@context` property from the unsecured doc.

dlongley (changes requested):

index.html
Wind4Greg: @dlongley Missed that. Fixed. Thanks.

index.html
```diff
@@ -892,11 +860,42 @@ <h4>Create Proof (eddsa-jcs-2022)</h4>
...
+Let |proof| be a clone of the proof options, |options|.
+              </li>
+              <li>
+Set <var>proof</var>.<var>@context</var> to
+<var>unsecuredDocument</var>.<var>@context</var>.
+              </li>
+              <li>
+Let |proofConfig| be the result of running the algorithm in
+Section [[[#proof-configuration-eddsa-jcs-2022]]] with
+|proof| passed as a parameter.
```
dlongley: This might be a bit more clear:

```suggestion
|proof| passed as the <em>proof options</em> parameter.
```

dlongley (approved):

msporny (approved):

msporny:

Issue #74, opened 2024-01-02
Ensure additional custom proof options provided via `proof` are included in the proof configuration


dlongley:
When creating a proof, other custom proof fields might be given, but it looks like the proof configuration algorithm will not include these -- and it should.

msporny:
This spec seems to be using the "old language" as well @Wind4Greg -- thoughts?

Wind4Greg:
@msporny, Let me get a PR out to fix on Monday.

msporny:
PR #88 has been merged to address this issue. Closing.

PR #88, opened 2024-07-01
Incorporate additional custom proof options

Wind4Greg:
This PR addresses issue https://github.com/w3c/vc-di-eddsa/issues/74. It updates the proof configuration procedures to use a clone of the options input and then updates the procedures accordingly.

***
<a href="https://pr-preview.s3.amazonaws.com/Wind4Greg/vc-di-eddsa/pull/88.html" title="Last updated on Jul 15, 2024, 4:02 PM UTC (a66c754)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/vc-di-eddsa/88/b41cbf7...Wind4Greg:a66c754.html" title="Last updated on Jul 15, 2024, 4:02 PM UTC (a66c754)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -480,7 +480,7 @@ Return |cryptosuite|.
 
         <p>
 The `eddsa-rdfc-2022` cryptographic suite takes an input document, canonicalizes
-the document using the RDF Dataset Canonicalization algorithm [[RDF-CANON]], and then 
+the document using the RDF Dataset Canonicalization algorithm [[RDF-CANON]], and then
 cryptographically hashes and signs the output
 resulting in the production of a data integrity proof. The algorithms in this
 section also include the verification of such a data integrity proof.
@@ -710,50 +710,33 @@ cryptographic suite</a> (<var>type</var>) and MUST contain a cryptosuite
 identifier (<var>cryptosuite</var>). A <em>proof configuration</em>
 object is produced as output.
           </p>
-
           <ol class="algorithm">
             <li>
-Let <var>proofConfig</var> be an empty object.
+Let |proofConfig| be a clone of the |options| object.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>type</var> to
-<var>options</var>.<var>type</var>.
-            </li>
-            <li>
-If <var>options</var>.<var>cryptosuite</var> is set, set
-<var>proofConfig</var>.<var>cryptosuite</var> to its value.
-            </li>
-            <li>
-If <var>options</var>.<var>type</var> is not set to `DataIntegrityProof` and
-<var>proofConfig</var>.<var>cryptosuite</var> is not set to `eddsa-rdfc-2022`,
-an error MUST be raised and SHOULD convey an error type of
+If |proofConfig|.|type| is not set to `DataIntegrityProof` and/or
+|proofConfig|.|cryptosuite| is not set to `eddsa-rdfc-2022`, an
+error MUST be raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>created</var> to
-<var>options</var>.<var>created</var>. If the value is not a valid
-[[XMLSCHEMA11-2]] datetime,
-an error MUST be raised and SHOULD convey an error type of
+If |proofConfig|.|created| is set and if the value is not a
+valid [[XMLSCHEMA11-2]] datetime, an error MUST be
+raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>verificationMethod</var> to
-<var>options</var>.<var>verificationMethod</var>.
+Set |proofConfig|.<var>@context </var>to
+|unsecuredDocument|.<var>@contex</var>t.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>proofPurpose</var> to
-<var>options</var>.<var>proofPurpose</var>.
+Let |canonicalProofConfig| be the result of applying the
+Universal RDF Dataset Canonicalization Algorithm
+[[RDF-CANON]] to the |proofConfig|.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>@context</var> to
-<var>unsecuredDocument</var>.<var>@context</var>
-            </li>
-            <li>
-Let <var>canonicalProofConfig</var> be the result of applying the
-RDF Dataset Canonicalization algorithm [[RDF-CANON]] to the <var>proofConfig</var>.
-            </li>
-            <li>
-Return <var>canonicalProofConfig</var>.
+Return |canonicalProofConfig|.
             </li>
           </ol>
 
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -710,50 +710,33 @@ <h4>Proof Configuration (eddsa-rdfc-2022)</h4>
...
+|proofConfig|.|cryptosuite| is not set to `eddsa-rdfc-2022`, an
+error MUST be raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>created</var> to
-<var>options</var>.<var>created</var>. If the value is not a valid
-[[XMLSCHEMA11-2]] datetime,
-an error MUST be raised and SHOULD convey an error type of
+If |proofConfig|.|created| is set and if the value is not a
```
```suggestion
If |proofConfig|.|created| is set to a value that is not a
```

index.html
```diff
@@ -710,50 +710,33 @@ <h4>Proof Configuration (eddsa-rdfc-2022)</h4>
...
+valid [[XMLSCHEMA11-2]] datetime, an error MUST be
+raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>verificationMethod</var> to
-<var>options</var>.<var>verificationMethod</var>.
+Set |proofConfig|.<var>@context </var>to
+|unsecuredDocument|.<var>@contex</var>t.
```
```suggestion
|unsecuredDocument|.<var>@context</var>.
```

index.html
```diff
@@ -710,50 +710,33 @@ <h4>Proof Configuration (eddsa-rdfc-2022)</h4>
...
             </li>
             <li>
-Set <var>proofConfig</var>.<var>proofPurpose</var> to
-<var>options</var>.<var>proofPurpose</var>.
+Let |canonicalProofConfig| be the result of applying the
+Universal RDF Dataset Canonicalization Algorithm
```
```suggestion
RDF Dataset Canonicalization Algorithm
```

index.html
```diff
@@ -1090,27 +1073,18 @@ <h4>Proof Configuration (eddsa-jcs-2022)</h4>
...
 <var>proofConfig</var>.<var>cryptosuite</var> is not set to `eddsa-jcs-2022`,
 an error MUST be raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>created</var> to
-<var>options</var>.<var>created</var>. If the value is not a valid
-[[XMLSCHEMA11-2]] datetime,
-an error MUST be raised and SHOULD convey an error type of
+If |proofConfig|.|created| is set and if the value is not a
```
```suggestion
If |proofConfig|.|created| is set to a value that is not a
```

index.html
```diff
@@ -1090,27 +1073,18 @@ <h4>Proof Configuration (eddsa-jcs-2022)</h4>
 Let <var>proofConfig</var> be a clone of the <var>options</var> object.
             </li>
             <li>
-If <var>options</var>.<var>type</var> is not set to `DataIntegrityProof` and
+If <var>proofConfig</var>.<var>type</var> is not set to `DataIntegrityProof` and
```
TallTed: Should this be `and/or` instead of `and`?
```suggestion
If <var>proofConfig</var>.<var>type</var> is not set to `DataIntegrityProof` and/or
```

index.html
```diff
@@ -1953,7 +1927,7 @@ <h3>Ed25519Signature2020</h3>
 
           <p>
   The `Ed25519Signature2020` cryptographic suite takes an input document,
-  canonicalizes the document using the RDF Dataset Canonicalization algorithm [[RDF-CANON]], 
+  canonicalizes the document using the RDF Dataset Canonicalization algorithm [[RDF-CANON]],
   and then cryptographically hashes and signs the output
```
```suggestion
  and then cryptographically hashes and signs the output,
```

index.html
```diff
@@ -2112,35 +2086,18 @@ <h4>Proof Configuration (Ed25519Signature2020)</h4>
...
-  <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
-              </li>
-              <li>
-  Set <var>proofConfig</var>.<var>verificationMethod</var> to
-  <var>options</var>.<var>verificationMethod</var>.
-              </li>
-              <li>
-  Set <var>proofConfig</var>.<var>proofPurpose</var> to
-  <var>options</var>.<var>proofPurpose</var>.
+If |proofConfig|.|created| is set and if the value is not a
```
```suggestion
If |proofConfig|.|created| is set to a value that is not a
```

TallTed (changes requested):
Editorial. Grammar, etc.

TallTed (approved):

index.html
```diff
@@ -710,50 +710,33 @@ <h4>Proof Configuration (eddsa-rdfc-2022)</h4>
...
+error MUST be raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>created</var> to
-<var>options</var>.<var>created</var>. If the value is not a valid
-[[XMLSCHEMA11-2]] datetime,
-an error MUST be raised and SHOULD convey an error type of
+If |proofConfig|.|created| is set to a value that is not a
+valid [[XMLSCHEMA11-2]] datetime, an error MUST be
```
dlongley: So `proofConfig.created` is optional and this doesn't read that way. If we have similar text in other cryptosuites we'll need to fix them as well. Maybe this works:

```suggestion
If |proofConfig|.|created| is present and set to a value that is not a
valid [[XMLSCHEMA11-2]] datetime, an error MUST be
```

index.html
```diff
@@ -710,50 +710,33 @@ <h4>Proof Configuration (eddsa-rdfc-2022)</h4>
...
+raised and SHOULD convey an error type of
 <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
             </li>
             <li>
-Set <var>proofConfig</var>.<var>verificationMethod</var> to
-<var>options</var>.<var>verificationMethod</var>.
+Set |proofConfig|.<var>@context </var>to
+|unsecuredDocument|.<var>@context</var>.
```
dlongley, line 721: This seems like it was erroneously copied from the jcs cryptosuite ... am I just misreading it? It looks like it might be in the wrong place because I didn't see it in the jcs cryptosuite in this spec.

index.html
```diff
@@ -2112,35 +2086,18 @@ <h4>Proof Configuration (Ed25519Signature2020)</h4>
...
-  <a data-cite="VC-DATA-INTEGRITY#PROOF_GENERATION_ERROR">PROOF_GENERATION_ERROR</a>.
-              </li>
-              <li>
-  Set <var>proofConfig</var>.<var>verificationMethod</var> to
-  <var>options</var>.<var>verificationMethod</var>.
-              </li>
-              <li>
-  Set <var>proofConfig</var>.<var>proofPurpose</var> to
-  <var>options</var>.<var>proofPurpose</var>.
+If |proofConfig|.|created| is set to a value that is not a
```
dlongley:
```suggestion
If |proofConfig|.|created| is present and set to a value that is not a
```

dlongley (changes requested):

index.html
```diff
@@ -1090,27 +1073,18 @@ <h4>Proof Configuration (eddsa-jcs-2022)</h4>
 Let <var>proofConfig</var> be a clone of the <var>options</var> object.
             </li>
             <li>
-If <var>options</var>.<var>type</var> is not set to `DataIntegrityProof` and
+If <var>proofConfig</var>.<var>type</var> is not set to `DataIntegrityProof` and/or
```
dlongley: Maybe this should just read "or"?

```suggestion
If <var>proofConfig</var>.<var>type</var> is not set to `DataIntegrityProof` or
```

index.html
Wind4Greg: I think this is right. Before we can canonize the proofConfig we need to add `@context`. Right? This was something that we had to fix way back when I started figuring out how to generate the test vectors...

index.html
Wind4Greg: @TallTed your opinion? I think this was your wording, but I can't remember.

TallTed: I prefer the accurate conjunction be used, which could be `or`, `and`, or `and/or`. I may not have always understood which was intended, so some of these conditionals may not now be correct.

* `And` (almost obviously) requires both conditions be met for the `then`.

* `Or` is (too often) treated as an `exclusive or`, meaning that _if either but not both conditions are met_, the `then` is active, but _if both conditions are met_, the `then` is _not_ active.

* `And/or` is usually read correctly to mean _if either or both conditions are met_, the `then` is active.

It might be better to change all of these phrasings to use the `either`, `both`, or `either or both` (possibly plus `neither`) conditionals to make clear the intended interpretation.

msporny:
@Wind4Greg merge conflicts, please fix. 

@dlongley and @Wind4Greg, I can't tell if we still need to make changes on this PR or not.

Wind4Greg:
@msporny  fixed merge conflicts as best I could. Don't know if I dropped some punctuation updates or not. A lot of small editorial updates have been happening.

index.html
dlongley: Ah, yes, ok.

dlongley (approved):

msporny (approved):

msporny:

